PR #46, opened 2026-04-27
fix: promote base/quote roles to first-class API parameters (#45)

### LINKED ISSUE #45: Add base/quote role assignment to pool search and detail endpoints
## Context

The Create Position flow's pool selection UI exposes Base Token and Quote Token fields, but these roles never reach the API. Search results come back in pool-native token0/token1 ordering, which on Mainnet renders WETH/USDC pools as "USDC/WETH" purely because USDC's address sorts lower than WETH's — on Arbitrum the same pair sorts the other way. The user-visible Flip button in the Summary compensates per-selection, but the role information is otherwise lost.

This will become important soon for downstream calculations where directionality matters. The recent sigma calculation narrowly avoided the issue only because log-returns are non-directional; future metrics will not be so forgiving.

## Goal

Promote base/quote role assignment from UI-local state to a first-class search and detail parameter so that:

- Search results can be filtered by the cartesian product of base and quote token sets.
- Each pool result carries an optional annotation indicating which side is quote, allowing consumers to render pairs in user-intended orientation regardless of pool-native token order.
- The pool detail endpoint accepts the same annotation directly.

Pool itself remains role-agnostic. This is purely a query-side concern.

## Current Behavior

- Search requires both Base and Quote token fields to be filled before firing.
- UI prevents the same token on both sides.
- Search returns pools in pool-native token0/token1 ordering.
- No role information is passed to or returned from the API.

### Pool Search Endpoint (breaking change)

Replace the current token parameter with two parameters:

- `base`: array of token addresses for the base side
- `quote`: array of token addresses for the quote side

Search returns the cartesian product `base × quote`: every pool whose token set is exactly one element from `base` and one from `quote`. Same-side pairs (e.g. WETH/frxETH when both are in `base`) are not returned.

Example: `base=[WETH, frxETH]&quote=[USDC, USDT]` returns pools matching {WETH/USDC, WETH/USDT, frxETH/USDC, frxETH/USDT}.

Each pool result is augmented with:

```ts
pool.userProvidedInfo = { isToken0Quote: boolean }
```

`isToken0Quote` is derived per result by checking which of the pool's token0/token1 appears in the `quote` array.

Minimum requirement: at least one token on each side (i.e. at least one pair in the cartesian product). See "Items to verify" below.

### Pool Detail Endpoint (`GET /pool/...`)

Add an optional query parameter:

- `isToken0Quote`: boolean

If present, the response includes `pool.userProvidedInfo = { isToken0Quote }`. If absent, the field is omitted (default = pool-native token0/token1 ordering).

This endpoint does not take `base` / `quote` arrays because the tokens are already determined by the pool address.

### Favorites

Favorites store the role assignment alongside the pool reference. When listing favorites, each entry carries its stored `userProvidedInfo`.

When a search result coincides with a favorited pool, the **active search parameters override** the favorite's stored role for that result. The Favorites view itself continues to display the stored role.

### Direct Address

Direct Address does not pre-assign roles. The result follows pool-native token0/token1 ordering (no `userProvidedInfo` field). The user can correct orientation via the existing Flip button in the Summary section.

## Items to Verify and Possibly Adjust

The following are current behaviors that should be re-examined as part of this work:

1. **Single-side empty parameter.** Currently the search only fires when both fields are populated. Verify whether allowing one side to be empty (interpreted as "any token on this side", paired with each token on the populated side) is desirable and feasible. Until decided, keep the current rule: at least one token on each side.
2. **Same token on both sides.** UI already prevents this. Verify that the API rejects this case explicitly (suggested: `400 Bad Request`) rather than silently filtering it out.

## Naming

Per-result annotation field name: `userProvidedInfo`.

Rationale: the data is contextual to a request — an echo of the user-supplied search parameters — not a stored user preference. `userPreferences` would imply persistent user-level settings, which this is not. The favorites case may look preference-like, but the field on a search result is still just an echo of the active query params (which override the favorite when the two disagree).

## Consumers to Migrate

The breaking API change requires migration in:

- Frontend UI (Create Position flow, Favorites tab, Direct Address tab)
- MCP server

## Out of Scope

- Persisting role assignment as a Pool property. Pool remains role-agnostic.
- Position model changes. `Position.isToken0Quote` already exists and is set at position-creation time from the user's chosen orientation; that flow is unchanged.

0xNedAlbo:
## Summary

- Promote base/quote role assignment from UI-local state to first-class API parameters. Pool search renames `tokenSetA`/`tokenSetB` → `base`/`quote`, and each result is annotated with `userProvidedInfo: { isToken0Quote }` derived per-chain from the request's quote set, so consumers can render pairs in user-intended orientation regardless of pool-native token order. Pool detail accepts an optional `isToken0Quote` query param it echoes back. Favorites persist orientation alongside the pool hash with lazy backwards-compat for legacy `string[]` entries.
- MCP `get_pool` gains the echo input; new `search_pools` tool surfaces the search endpoint with an exact-symbol-or-address contract documented in the tool's input schema.
- UI: `PoolTable` now pivots the pair label by `userProvidedInfo`; the wizard `SELECT_POOL` reducer reads the orientation; `handleToggleFavorite` forwards the active orientation when persisting a favorite.
- Adds a project-local `.mcp.json` midcurve entry (env-var-interpolated key) so anyone working on this repo overrides their global install with the local dev API.

Fixes #45

## Test plan

- [ ] `pnpm typecheck` clean across all 14 packages
- [ ] api-shared schema tests: `pnpm --filter @midcurve/api-shared test:run` — 8 tests cover trivial-case rejection, case-insensitive symbol compare, EIP-55 normalize, `[WETH,stETH]×[WETH,stETH]` passes, defaults applied
- [ ] services tests: `pnpm --filter @midcurve/services test:run` — 16 new UserSettings tests cover legacy-string/object/mixed normalization, lazy upgrade on re-add, idempotent removes
- [ ] mcp-server tests: `pnpm --filter @midcurve/mcp-server test:run` — 8 new tests cover `formatPool` userProvidedInfo echo and `formatPoolSearchResult`
- [ ] API smoke (against local dev):
      ```bash
      # Search returns userProvidedInfo per result, with per-chain orientation
      curl -s -X POST http://localhost:3001/api/v1/pools/uniswapv3/search \
        -H 'content-type: application/json' \
        -d '{"base":["WETH"],"quote":["USDC"],"chainIds":[1,8453]}' \
        | jq '.data[] | {chain: .chainName, pair: "\(.token0.symbol)/\(.token1.symbol)", isToken0Quote: .userProvidedInfo.isToken0Quote}'
      # Mainnet pools → isToken0Quote=true (USDC is token0); Base pools → false (WETH is token0)

      # Trivial-case rejection
      curl -s -X POST http://localhost:3001/api/v1/pools/uniswapv3/search \
        -H 'content-type: application/json' \
        -d '{"base":["WETH"],"quote":["WETH"],"chainIds":[1]}' -i | head -1
      # → HTTP/1.1 400

      # Detail endpoint echoes the role
      curl -s 'http://localhost:3001/api/v1/pools/uniswapv3/1/0x88e6A0c2dDD26FEEb64F039a2c41296FcB3f5640?isToken0Quote=true' \
        | jq '.data.userProvidedInfo'
      ```
- [ ] MCP smoke (verified via Claude Code against dev server): `search_pools` annotated each result with the correct per-chain `isToken0Quote`; `get_pool` with `isToken0Quote=true` echoed the field at the response root
- [ ] UI smoke (verified via Chrome DevTools MCP): wizard pool table now displays user-intended `base / quote` orientation across Mainnet (where USDC is token0) and Base (where WETH is token0); Direct-Address tab still uses pool-native ordering
- [ ] Favorites round-trip: starring a pool from the search tab persists `isToken0Quote`; the Favorites tab returns it; a legacy `favoritePoolHashes: string[]` row is normalized on read with `isToken0Quote` undefined

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -14,6 +14,15 @@
         "--browser-url=http://127.0.0.1:9222",
         "--no-usage-statistics"
       ]
+    },
+    "midcurve": {
+      "type": "stdio",
+      "command": "node",
+      "args": ["apps/midcurve-mcp-server/dist/index.js"],
+      "env": {
+        "MIDCURVE_API_KEY": "${MIDCURVE_DEV_API_KEY}",
+        "MIDCURVE_API_URL": "http://localhost:3001"
+      }
     }
   }
 }
--- a/apps/midcurve-api/src/app/api/v1/pools/favorites/route.ts
+++ b/apps/midcurve-api/src/app/api/v1/pools/favorites/route.ts
@@ -88,7 +88,7 @@ export async function POST(request: NextRequest): Promise<Response> {
         });
       }
 
-      const { protocol, chainId, poolAddress } = validation.data;
+      const { protocol, chainId, poolAddress, isToken0Quote } = validation.data;
 
       // 2. Delegate to protocol-specific service
       apiLog.businessOperation(
@@ -97,7 +97,7 @@ export async function POST(request: NextRequest): Promise<Response> {
         'addFavorite',
         'FavoritePool',
         `${protocol}/${chainId}/${poolAddress}`,
-        { protocol, chainId, poolAddress }
+        { protocol, chainId, poolAddress, isToken0Quote }
       );
 
       let result;
@@ -110,6 +110,7 @@ export async function POST(request: NextRequest): Promise<Response> {
               userId: user.id,
               chainId,
               poolAddress,
+              ...(typeof isToken0Quote === 'boolean' && { isToken0Quote }),
             });
           } catch (error) {
             if (error instanceof Error) {
@@ -221,6 +222,9 @@ export async function POST(request: NextRequest): Promise<Response> {
         poolAddress: result.pool.address,
         pool: serializedPool as unknown as FavoritePoolItem['pool'],
         metrics: metricsBlock,
+        ...(typeof result.isToken0Quote === 'boolean' && {
+          userProvidedInfo: { isToken0Quote: result.isToken0Quote },
+        }),
       };
 
       const responseData: AddFavoritePoolData = {
@@ -368,6 +372,9 @@ export async function GET(request: NextRequest): Promise<Response> {
           poolAddress: fav.pool.address,
           pool: serializedPool as unknown as FavoritePoolItem['pool'],
           metrics: buildPoolMetricsBlock(subgraphInput, sigma),
+          ...(typeof fav.isToken0Quote === 'boolean' && {
+            userProvidedInfo: { isToken0Quote: fav.isToken0Quote },
+          }),
         });
       }
 
--- a/apps/midcurve-api/src/app/api/v1/pools/uniswapv3/[chainId]/[address]/route.ts
+++ b/apps/midcurve-api/src/app/api/v1/pools/uniswapv3/[chainId]/[address]/route.ts
@@ -48,6 +48,9 @@ export async function OPTIONS(request: NextRequest): Promise<Response> {
  * Query params:
  * - metrics (optional): Include subgraph metrics (TVL, volume, fees). Defaults to false.
  * - fees (optional): Include fee data for APR calculations (24h volumes, token prices). Defaults to false.
+ * - isToken0Quote (optional): If provided, response includes
+ *   `userProvidedInfo: { isToken0Quote }` (pure echo — metrics/feeData are
+ *   not reoriented).
  */
 export async function GET(
   request: NextRequest,
@@ -102,7 +105,7 @@ export async function GET(
         });
       }
 
-      const { metrics, fees } = queryResult.data;
+      const { metrics, fees, isToken0Quote } = queryResult.data;
 
       // 3. Discover pool (creates if new, refreshes if existing)
       // This ensures we always return fresh on-chain state
@@ -240,6 +243,9 @@ export async function GET(
         pool: serializedPool as unknown as UniswapV3Pool,
         ...(metricsData && { metrics: metricsData }),
         ...(feeData && { feeData }),
+        ...(typeof isToken0Quote === 'boolean' && {
+          userProvidedInfo: { isToken0Quote },
+        }),
       };
 
       const response = createSuccessResponse(responseData, {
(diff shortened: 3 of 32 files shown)


0xNedAlbo:
Reviewed the diff against the agreed plan. Implementation is high-quality — schema validation, per-chain quote-set, self-exclusion logic, JSDoc contract, favorite-toggle wiring, reducer + display orientation, and lazy backwards-compat for favorites are all faithful to spec. Schema tests (8) and UserSettings tests (16) are thorough.

**One gap before merging: missing service-layer tests for `uniswapv3-pool-search-service`.**

The plan called for two service tests that don't appear in the diff:

1. **Per-chain orientation derivation** — given a multi-chain query where the same symbol resolves to different addresses across chains (e.g., USDC on Mainnet vs USDC on Base, where USDC sorts as token0 on Mainnet but token1 on Base), verify each result gets the correct per-chain `isToken0Quote`. This exercises the new `Map<chainId, Set<address>>` and is the heart of the per-chain refinement.

2. **Self-exclusion at the service layer** — call `searchPools()` directly with `base=["WETH"], quote=["WETH"]` (bypassing the route's schema validation). Expect `[]`, not a thrown error. This pins the contract that the trivial case is handled gracefully even when schema is bypassed (e.g., from another service caller in the future).

Schema-level and storage-layer tests are excellent, but the service-layer logic where the per-chain map is built and self-exclusion happens is currently only verified via manual smoke. Adding these two tests closes the loop.

**Minor observations (not blockers):**

- `.mcp.json` change is out of scope for #45 — useful dev tooling, but technically separate. Consider splitting into its own commit/PR if you want a clean #45 history.
- "Lazy backwards-compat" is in practice "lazy on read, eager on next write": `addFavoritePoolEntry` reads-normalizes-upserts, so any favorite touch upgrades all of that user's entries to the new shape. Functionally fine (arguably better than strict-lazy), just worth knowing it doesn't strictly preserve legacy strings forever.
- `useDiscoverPool` forwards `isToken0Quote` and the response echoes `userProvidedInfo`, but the SELECT_POOL flow uses the original `PoolSearchResultItem.userProvidedInfo` as source-of-truth — the discovery echo is unused. Harmless but redundant; could be removed if you want a tighter surface.

PR #55, opened 2026-04-27
feat: LVR-Coverage indicator + column manager in pool search table

### LINKED ISSUE #54: RFC-0002: LVR-Coverage indicator in the pool search table (frontend)
**Status:** implemented
**Date:** 2026-04-27
**Slug:** lvr-coverage-frontend
**Implementation:** #55

## Summary

Surfaces the LVR-Coverage indicator from RFC-0001 in the pool search table (`apps/midcurve-ui/src/components/positions/wizard/create-position/uniswapv3/shared/PoolTable.tsx`) as a 5-band visual pill, alongside a column-management mechanism that lets users toggle which metric columns are visible. Default-visible set is intentionally minimal (Pool · TVL · Fee APR 7d · LVR-Coverage); everything else is opt-in through a gear-icon popover. User selection persists server-side via `UserSettingsData`.

## Context

RFC-0001 added `coverageLongTerm` (raw ratio) and `coverageBand` (5-band enum) to `SigmaFilterBlock`. The backend now ships the canonical mapping; the frontend currently does nothing with it.

Today's pool search table shows: Pool · TVL · Volume 24h · Fees 24h · ø APR 7d (with the existing yellow `!` heuristic). The yellow `!` is purely client-side and signals data-quality concerns ("don't trust the APR if TVL/fees/volume look wrong"). It is orthogonal to the LVR signal — `!` is about whether the inputs are usable, the LVR pill is the economic verdict given those inputs.

Goal of this RFC: turn the table from a generic activity dashboard into a decision tool for "is this pool a viable LP target". Concretely: the user scans left to right and reads "WETH/USDC on Base, $126M, 49.7% APR, deep_green pill 7.8×" and has enough to mark it as a candidate.

### Column model

Three categories:

**Always-visible (not in the column manager):**
- `pool` — identifier (token pair, chain, fee tier). Never hidden, never reordered.
- `favorite` (the star) — per-row action, not a data column. Preserved exactly as currently implemented in `PoolTable.tsx`. Out of scope for any change in this RFC.

**Default-visible toggleable:**
- `tvl` — Total Value Locked, USD.
- `feeApr7d` — fee APR over the 7-day window, as percent.
- `lvrCoverage` — the new 5-band pill (see below).

**Optional (default-hidden, available in the column manager):**
- `volume7dAvg` — average daily volume over 7 days, USD.
- `fees24h` — fees collected last 24h, USD.
- `lvrThreshold` — `sigmaSqOver8_365d` as percent (the σ²/8 LVR threshold).
- `margin` — `marginLongTerm` as signed percent.
- `coverageRatio` — `coverageLongTerm` as a numeric ratio (the raw number behind the pill, for users who want the figure without the band visualization).
- `sigmaPair365d` — the synthetic cross-pair σ over 365d, as percent.
- `velocity` — `sigma60d / sigma365d` ratio.
- `verdict60d` — short-term verdict (PASS / FAIL / INSUFFICIENT_DATA).
- `verdictAgreement` — AGREE / DIVERGENT / INSUFFICIENT_DATA between long and short term.

Removed from default-visible (vs. today): Volume 24h and Fees 24h. Within a single fee tier, both are functions of each other and of TVL × APR — they don't add decision-relevant information at default-scan time. Available on demand through the manager.

### LVR-Coverage pill

Rendered in the `lvrCoverage` cell. Single rounded pill, colored by `coverageBand`, text inside the pill = the coverage value.

| `coverageBand` | Pill color | Text inside |
|---|---|---|
| `deep_red` | dark red (e.g. `bg-red-700`) | coverage value |
| `red` | red (e.g. `bg-red-500`) | coverage value |
| `yellow` | yellow (e.g. `bg-yellow-500`) | coverage value |
| `green` | green (e.g. `bg-green-500`) | coverage value |
| `deep_green` | dark green (e.g. `bg-green-700`) | coverage value |
| `insufficient_data` | gray (e.g. `bg-gray-500`) | `n/a` |

Color tokens to follow whatever design system the codebase already uses (likely Tailwind classes; align with existing `text-yellow-400` / `text-green-400` patterns in `PoolTable.tsx`).

**Coverage value formatting:**
- Single decimal place: `2.4×`, `0.7×`, `0.4×` (using the `×` character, not `x`).
- Cap: any value `≥ 5.0` displays as `>5×` regardless of precision. Cap is purely display logic — backend ships the raw `coverageLongTerm` ratio (RFC-0001 was explicit on this).
- `null` coverage (i.e. `coverageBand === 'insufficient_data'`): pill displays `n/a`.

**Tooltip on hover:**
- For data-bearing pills (any non-`insufficient_data`):
  - `LVR threshold (σ²/8, 365d): 6.4%` — derived from `sigmaSqOver8_365d`.
  - `Coverage ratio: 7.81` — the un-rounded `coverageLongTerm` value.
  - `Fee APR (7d): 49.7%` — for context.
- For `insufficient_data` pills, derive the reason from the underlying `volatility` block, in this priority order:
  - If either `volatility.token0.sigma365d.status === 'token_not_listed'` or `.token1.sigma365d.status === 'token_not_listed'`: `{symbol} not listed on CoinGecko` — name the offending token by symbol.
  - Else if either token's `sigma365d.status === 'insufficient_history'` or `volatility.pair.sigma365d.status === 'insufficient_history'`: `Pool or token too young for σ calculation`.
  - Else if `fetch_failed`: `Volatility data temporarily unavailable — try again later`.
  - Else: generic `LVR coverage cannot be computed for this pool`.

### Relationship to the existing `!` warning

Unchanged. `shouldShowAprWarning` (PoolTable.tsx:50-62) keeps firing on its current heuristic (TVL=0 / fees=0 / volume=0 with non-zero APR). The `!` icon stays in the APR cell as a small inline marker. The LVR pill in its own cell is independent — the two indicators answer different questions and may both fire for the same row.

### Column manager

A gear icon (e.g. `lucide-react`'s `Settings` or `Cog`) rendered to the right of the "Available Pools" section header.

Click opens a popover (not a modal — modal is too heavy for a frequent toggle action). Popover contents:

- Section header: "Visible columns".
- Checkbox list of all toggleable columns (default + optional, in the same hardcoded display order as the table itself). Star and Pool are not in the list; they are not toggleable.
- Each row: `[checkbox] [column display name]`.
- Optional section header: "Default columns" / "Additional columns" — keeps the popover scannable. If purely cosmetic, can be omitted.
- No drag-and-drop reordering. Display order is hardcoded in component code; the user only controls visibility. Reordering is YAGNI for v1.
- No "save" button. Toggles persist immediately (debounced if necessary).
- Click outside or Escape closes the popover.

### Persistence

New top-level field on `UserSettingsData` (typed in `packages/midcurve-shared/src/types/user-settings.ts`):

```typescript
export type PoolTableColumnId =
  | 'tvl'
  | 'feeApr7d'
  | 'lvrCoverage'
  | 'volume7dAvg'
  | 'fees24h'
  | 'lvrThreshold'
  | 'margin'
  | 'coverageRatio'
  | 'sigmaPair365d'
  | 'velocity'
  | 'verdict60d'
  | 'verdictAgreement';

export interface UserSettingsData {
  favoritePoolHashes: FavoritePoolEntry[];
  costBasisMethod: CostBasisMethod;
  poolTableVisibleColumns: PoolTableColumnId[]; // NEW
}
```

Default value in `DEFAULT_USER_SETTINGS`:

```typescript
poolTableVisibleColumns: ['tvl', 'feeApr7d', 'lvrCoverage'],
```

**Lazy backwards compatibility (mirrors the `favoritePoolHashes` legacy pattern already in this file):** when reading a `UserSettings` row that predates this field, the read layer fills in the default. Writes always emit the explicit array. No Prisma migration needed — the column is `Json`, schema is unchanged.

### Sorting

The API endpoint `POST /api/v1/pools/uniswapv3/search` accepts `sortBy` only for: `tvlUSD`, `volume24hUSD`, `fees24hUSD`, `volume7dAvgUSD`, `fees7dAvgUSD`, `apr7d`.

Sort behavior in the table:
- Sortable columns (clickable header with sort arrow): `tvl`, `feeApr7d`, plus the optional `volume7dAvg` and `fees24h` when visible.
- Non-sortable columns (no sort arrow rendered, header is plain text): `lvrCoverage`, `lvrThreshold`, `margin`, `coverageRatio`, `sigmaPair365d`, `velocity`, `verdict60d`, `verdictAgreement`.

Server-side sort support for the new metric columns is out of scope for this RFC. If demand emerges, file a separate RFC for the API extension.

Default sort: TVL descending (matches current behavior).

## Alternatives considered

- **5-band continuous gradient bar instead of discrete pill.** Rejected. The asymmetric cutoffs in RFC-0001 (yellow stretches 0.9–1.5, deep_red below 0.5) carry meaning that a continuous gradient flattens. Discrete pill preserves the categorical bands as the user sees them.
- **LVR threshold as default-visible column instead of tooltip.** Considered, rejected. Adds a sixth column to the default view, increases scan width, and the threshold is rarely the question being asked at scan time. Available as opt-in column for users who want it.
- **Per-band `coverageBand` rendered as text label (`PASS` / `FAIL` / `MARGINAL`) instead of pill.** Rejected. The 5-band scheme doesn't map cleanly to PASS/FAIL — yellow is neither — and a text label loses the visual scanability that's the whole point of the indicator.
- **Generic `tableSettings` container in `UserSettingsData` (e.g. `tableSettings: { pools: { visibleColumns: ... }, positions: { ... } }`).** Rejected. Spec'ing for the position table now is YAGNI; when the position table needs the same treatment, add a sibling top-level field then. The cost of adding a second top-level field later is one TypeScript line; the cost of carrying nested infrastructure now is real.
- **Column drag-and-drop reordering.** Deferred. Adds non-trivial UI infrastructure (react-dnd or similar), and the hardcoded display order is sensible for the scan workflow we're optimizing for.
- **Per-row LVR breakdown popover (e.g. click on pill to see σ_token0, σ_token1, σ_pair, fee APR, threshold all at once).** Out of scope. The hover tooltip carries the essential numbers; a deeper breakdown would belong to a future "compare pools" or "pool detail" view, not the search table.

## Open Questions

_None — concept-phase decisions are settled._

## Out of Scope

- Drag-and-drop column reordering.
- Mobile-specific responsive layout. Today's table is desktop-oriented; mobile is a separate UX concern that would warrant its own RFC.
- Server-side sorting for the new metric columns (`lvrCoverage`, `margin`, `coverageRatio`, `sigmaPair365d`, etc.). Requires API extension; not blocking.
- "Compare pools" detail view. Separate future feature.
- Position table column manager. When that need surfaces, it gets its own RFC and a parallel `positionTableVisibleColumns` field on `UserSettingsData`.
- Frontend caching strategy for repeated `UserSettings` writes from rapid toggling. If toggle latency becomes a UX issue, debounce on the client side; out of scope as a contractual concern.
- Changes to the existing `shouldShowAprWarning` heuristic. The yellow `!` stays as-is; LVR pill is orthogonal.

## Decision provenance

All decisions in this draft were confirmed in the concept chat on 2026-04-27 (continuation of the RFC-0001 session):

- Default-visible set (Pool · TVL · Fee APR 7d · LVR-Coverage, plus always-visible Star action): user-confirmed; favorite star is preserved exactly as currently implemented and explicitly out of scope for any change here.
- Volume 24h / Fees 24h removed from default: user reasoning ("Fees vs TVL ist eigentlich spannend; Volume ist Funktion davon"), assistant elaborated on Volume/TVL turnover redundancy with the LVR pill, user confirmed.
- LVR threshold as hover tooltip, not default column: user-confirmed.
- 5-band pill with coverage value as text inside the pill: assistant-proposed (dual information — color = band, number = precise value), user-confirmed.
- Cap logic (`>5×`) lives in the UI, not the backend: chat-confirmed in RFC-0001 context, restated here.
- Insufficient_data pill displays `n/a` and derives reason from `volatility.*.sigmaXXXd.status`: assistant-proposed mapping to `SigmaStatus` values, user-confirmed earlier in the session ("wir sollten da in grau eine Art Begründung liefern 'Pool too young' etc.").
- Column manager as gear icon (option b from the earlier three-way choice): user-confirmed.
- Server-side persistence via `UserSettings`, specific `poolTableVisibleColumns` field at top level (not nested `tableSettings`): user-confirmed ("lieber spezifisch: `poolTableVisibleColumns: PoolTableColumnId[]` direkt auf Top-Level. Einfach, lokal, klar.").
- Lazy backwards-compat for missing field (mirrors `favoritePoolHashes` legacy pattern): assistant-proposed, consistent with existing schema convention.
- Non-sortable columns show no sort arrow: chat-confirmed; server-side sort extension parked as separate concern.
- Mobile out of scope: chat-confirmed.
- Star column unchanged from current implementation: user-confirmed in this turn ("wir haben diese Spalte schon drin. Sie muss nur unangetastet bleiben und immer sichtbar sein.").

0xNedAlbo:
## Summary

- Adds the RFC-0002 frontend treatment on top of RFC-0001's backend signals: 5-band LVR-Coverage pill with hover tooltip in the pool search table, plus a gear-icon column manager that lets users toggle which metric columns are visible.
- New persisted preference `poolTableVisibleColumns` on `UserSettingsData` (defaults to `Pool · TVL · Fee APR 7d · LVR-Coverage`); read/write via a dedicated `GET`/`PUT` at `/api/v1/user/me/settings/pool-table-columns` with optimistic-update mutation on the client.
- Lazy back-compat: legacy `UserSettings` rows automatically fall through to defaults via `UserSettingsService.getByUserId`. No Prisma migration.

## Test plan

- [x] `pnpm typecheck` (all 14 packages)
- [x] `pnpm --filter @midcurve/services test:run -- user-settings` (138 tests pass; new normalization + update tests cover defaults, unknown ids, dedup, empty arrays, round-trip)
- [ ] `pnpm dev`: in the create-position wizard, search for pools and verify
  - Default columns are Star · Pool · TVL · Fee APR 7d · LVR-Coverage
  - LVR-Coverage pill renders the band color and coverage value (`>5×` cap on rounded display)
  - Hovering a colored pill shows LVR threshold / Coverage ratio / Fee APR; gray pill shows derived reason
  - Gear icon opens popover with `Default columns` and `Additional columns` sections
  - Toggling columns updates the table immediately and persists across reload
  - The existing yellow `!` warning still fires alongside the LVR pill on degenerate-data pools

Fixes #54

### DIFF
--- a/apps/midcurve-ui/src/components/positions/wizard/create-position/uniswapv3/shared/PoolTable.tsx
+++ b/apps/midcurve-ui/src/components/positions/wizard/create-position/uniswapv3/shared/PoolTable.tsx
@@ -1,9 +1,19 @@
-import { useState } from 'react';
+import { useState, type ReactNode } from 'react';
 import { Star, ChevronUp, ChevronDown } from 'lucide-react';
 import type { PoolSearchResultItem } from '@midcurve/api-shared';
+import type { PoolTableColumnId } from '@midcurve/shared';
+import { LvrCoveragePill } from './LvrCoveragePill';
 
-type SortColumn = 'tvlUSD' | 'volume24hUSD' | 'fees24hUSD' | 'apr7d';
+type SortColumn = 'tvlUSD' | 'volume24hUSD' | 'fees24hUSD' | 'apr7d' | 'volume7dAvgUSD';
 type SortDirection = 'asc' | 'desc';
+
+interface ColumnDef {
+  id: PoolTableColumnId;
+  label: string;
+  align: 'left' | 'right';
+  sortKey?: SortColumn;
+  render: (pool: PoolSearchResultItem) => ReactNode;
+}
 
 interface PoolTableProps {
   pools: PoolSearchResultItem[];
@@ -11,7 +21,182 @@ interface PoolTableProps {
   onSelectPool: (pool: PoolSearchResultItem) => void;
   onToggleFavorite?: (pool: PoolSearchResultItem) => void;
   isLoading?: boolean;
+  visibleColumns: PoolTableColumnId[];
 }
+
+const formatCurrency = (value: string | number) => {
+  const numValue = typeof value === 'string' ? parseFloat(value) : value;
+  if (numValue >= 1_000_000) {
+    return `$${(numValue / 1_000_000).toFixed(1)}M`;
+  }
+  if (numValue >= 1_000) {
+    return `$${(numValue / 1_000).toFixed(1)}K`;
+  }
+  return `$${numValue.toFixed(0)}`;
+};
+
+const formatPercent = (value: number) => `${value.toFixed(1)}%`;
+
+const formatSignedPercent = (value: number | null) => {
+  if (value === null) return 'n/a';
+  const pct = value * 100;
+  const sign = pct > 0 ? '+' : '';
+  return `${sign}${pct.toFixed(1)}%`;
+};
+
+const formatRatio = (value: number | null, digits = 2) => {
+  if (value === null) return 'n/a';
+  return value.toFixed(digits);
+};
+
+const formatRawPercent = (value: number | null) => {
+  if (value === null) return 'n/a';
+  return `${(value * 100).toFixed(1)}%`;
+};
+
+const formatFeeTier = (feeTier: number) => `${(feeTier / 10000).toFixed(2)}%`;
+
+/**
+ * APR may be unreliable when pool has low/no liquidity, volume, or fees.
+ */
+const shouldShowAprWarning = (pool: PoolSearchResultItem): boolean => {
+  const tvl = parseFloat(pool.metrics.tvlUSD) || 0;
+  const volume = parseFloat(pool.metrics.volume24hUSD) || 0;
+  const fees = parseFloat(pool.metrics.fees24hUSD) || 0;
+  const apr = pool.metrics.apr7d || 0;
+
+  if (apr > 0 && tvl === 0) return true;
+  if (apr > 0 && fees === 0) return true;
+  if (volume === 0) return true;
+
+  return false;
+};
+
+/**
+ * Column registry. Display order is the array order; visibility is controlled
+ * by the `visibleColumns` prop.
+ */
+const COLUMNS: ColumnDef[] = [
+  {
+    id: 'tvl',
+    label: 'TVL',
+    align: 'right',
+    sortKey: 'tvlUSD',
+    render: (pool) => formatCurrency(pool.metrics.tvlUSD),
+  },
+  {
+    id: 'feeApr7d',
+    label: 'ø APR 7d',
+    align: 'right',
+    sortKey: 'apr7d',
+    render: (pool) => (
+      <div className="flex items-center justify-end gap-1">
+        <span className="text-green-400">{formatPercent(pool.metrics.apr7d)}</span>
+        {shouldShowAprWarning(pool) && (
+          <span
+            className="text-yellow-400 text-xs font-bold cursor-help"
+            title="APR may be unreliable: low TVL, volume, or fees"
+          >
+            !
+          </span>
+        )}
+      </div>
+    ),
+  },
+  {
+    id: 'lvrCoverage',
+    label: 'LVR-Coverage',
+    align: 'right',
+    render: (pool) => (
+      <div className="flex justify-end">
+        <LvrCoveragePill pool={pool} />
+      </div>
+    ),
+  },
+  {
+    id: 'volume7dAvg',
+    label: 'Volume 7d avg',
+    align: 'right',
+    sortKey: 'volume7dAvgUSD',
+    render: (pool) => formatCurrency(pool.metrics.volume7dAvgUSD),
+  },
+  {
+    id: 'fees24h',
+    label: 'Fees 24h',
+    align: 'right',
+    sortKey: 'fees24hUSD',
+    render: (pool) => formatCurrency(pool.metrics.fees24hUSD),
+  },
+  {
+    id: 'lvrThreshold',
+    label: 'LVR threshold',
+    align: 'right',
+    render: (pool) => formatRawPercent(pool.metrics.sigmaFilter.sigmaSqOver8_365d),
+  },
+  {
+    id: 'margin',
+    label: 'Margin',
+    align: 'right',
+    render: (pool) => {
+      const v = pool.metrics.sigmaFilter.marginLongTerm;
+      if (v === null) return <span className="text-slate-500">n/a</span>;
+      const pct = v * 100;
+      const cls = pct >= 0 ? 'text-green-400' : 'text-red-400';
+      return <span className={cls}>{formatSignedPercent(v)}</span>;
+    },
+  },
+  {
+    id: 'coverageRatio',
+    label: 'Coverage ratio',
+    align: 'right',
+    render: (pool) => formatRatio(pool.metrics.sigmaFilter.coverageLongTerm),
+  },
+  {
+    id: 'sigmaPair365d',
+    label: 'σ pair (365d)',
+    align: 'right',
+    render: (pool) => {
+      const v = pool.metrics.volatility.pair.sigma365d.value;
+      return formatRawPercent(v ?? null);
+    },
+  },
+  {
+    id: 'velocity',
+    label: 'Velocity',
+    align: 'right',
+    render: (pool) => formatRatio(pool.metrics.volatility.velocity),
+  },
+  {
+    id: 'verdict60d',
+    label: 'Verdict 60d',
+    align: 'right',
+    render: (pool) => {
+      const v = pool.metrics.sigmaFilter.verdictShortTerm;
+      const cls =
+        v === 'PASS'
+          ? 'text-green-400'
+          : v === 'FAIL'
+            ? 'text-red-400'
+            : 'text-slate-500';
+      return <span className={cls}>{v === 'INSUFFICIENT_DATA' ? 'n/a' : v}</span>;
+    },
+  },
+  {
+    id: 'verdictAgreement',
+    label: 'Agreement',
+    align: 'right',
+    render: (pool) => {
+      const v = pool.metrics.sigmaFilter.verdictAgreement;
+      const cls =
+        v === 'AGREE'
+          ? 'text-green-400'
+          : v === 'DIVERGENT'
+            ? 'text-yellow-400'
+            : 'text-slate-500';
+      return <span className={cls}>{v === 'INSUFFICIENT_DATA' ? 'n/a' : v}</span>;
+    },
+  },
+];
 
 export function PoolTable({
   pools,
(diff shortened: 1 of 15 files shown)


apps/midcurve-ui/src/components/positions/wizard/create-position/uniswapv3/shared/PoolTableColumnManager.tsx
```diff
@@ -0,0 +1,137 @@
...
+        type="button"
+        onClick={() => setIsOpen((v) => !v)}
+        className="p-1.5 rounded-md text-slate-400 hover:text-slate-200 hover:bg-slate-700/50 cursor-pointer transition-colors"
+        title="Manage visible columns"
+        aria-label="Manage visible columns"
+        aria-expanded={isOpen}
+      >
+        <Settings className="w-4 h-4" />
+      </button>
+
```
0xNedAlbo, line 141: **Bug: Column manager popover renders behind the pool list.**

When the gear icon is clicked, the popover opens at `z-30` relative to the `PoolTableColumnManager` parent (`relative inline-block`), but the table body that follows in the DOM ends up rendering on top of it. Likely the parent `PoolTableSection`'s header container has no explicit positioning / stacking context, so the popover's `z-30` is competing with the natural source order of the table body that comes after it.

**Two possible fixes:**

1. *Quick fix:* in `PoolTableSection.tsx`, give the header row container `relative z-20` (or similar): `<div className="flex items-center justify-between mb-4 relative z-20">`. This lifts the entire header (including the popover-anchoring gear button) above the table body in the stacking context.

2. *Robust fix (preferred long-term):* render the popover as a React portal to `document.body`. This escapes any parent stacking context entirely and is the standard pattern for popovers/tooltips/dropdowns. Headless UI / Radix do this by default. Worth it if more popovers appear in the codebase later; for a single-popover situation, fix (1) is fine.

Either fix solves the immediate bug. Pick whichever fits the codebase style — if there are no other portals in use yet, I'd go with (1) for simplicity.

apps/midcurve-ui/src/components/positions/wizard/create-position/uniswapv3/shared/PoolTable.tsx
```diff
@@ -154,18 +266,55 @@ export function PoolTable({
 
   return (
     <div className="h-full flex flex-col">
-      <h3 className="text-lg font-semibold text-white mb-4">Available Pools</h3>
-
       <div className="flex-1 overflow-auto">
         <table className="w-full text-left">
```
0xNedAlbo, line 270: **Bug: LVR-Coverage pill scrolls over the sticky column header.**

When the user scrolls the pool list, the colored pills in the `lvrCoverage` cells render *over* the sticky `<thead>` row instead of being clipped beneath it (visible in UI testing — the `1.3×` pill ends up obscuring the "LVR-Coverage" header text).

Root cause: the `<thead>` is `sticky top-0` with `bg-slate-800/90` (semi-transparent), but has no explicit `z-index`. Each `LvrCoveragePill` lives inside a `<div className="relative inline-block group">` which creates its own stacking context per cell, and without a higher z-index on the header those cells end up "in front" of it during scroll.

**Fix:** add `z-10` to the sticky `<thead>`:

```tsx
<thead className="sticky top-0 z-10 bg-slate-800/90 backdrop-blur-sm">
```

Alternatively (Tailwind allows this on the `<tr>` instead if needed), the same `z-10` on the header `<tr>` works. The header should always paint above scrolling cell content.

The semi-transparent background (`bg-slate-800/90`) is fine to keep — that's a deliberate design choice for the backdrop blur — once the z-index is set, the visual layering is correct even with transparency.

0xNedAlbo (commented):
Two visual bugs surfaced via UI testing — both CSS-only fixes (stacking-context / z-index issues). The data flow, component logic, persistence, and tooltip rendering all look correct. Inline comments below.

Issue #11, opened 2026-01-17
Bug: Same EVM address signing on different chains creates duplicate user accounts


0xNedAlbo:
## Bug Description

When a user signs a SIWE (Sign-In With Ethereum) message with the same EVM address but on different chains (e.g., Ethereum vs Arbitrum), the system creates **separate user accounts** instead of recognizing them as the same user.

## Expected Behavior

The same EVM address should be recognized as the same user regardless of which chain they sign on:

| Step | User Action | Expected Result |
|------|-------------|-----------------|
| 1 | Signs SIWE with `0xABC...` on Ethereum | User A created |
| 2 | Signs SIWE with `0xABC...` on Arbitrum | **Same User A returned** ✅ |

## Actual Behavior

| Step | User Action | Actual Result |
|------|-------------|---------------|
| 1 | Signs SIWE with `0xABC...` on Ethereum | User A created |
| 2 | Signs SIWE with `0xABC...` on Arbitrum | **User B created** ❌ |

## Root Cause

The `AuthWalletAddress` table uses a composite unique constraint `@@unique([address, chainId])`, and the lookup in the verify endpoint uses both fields:

```typescript
// apps/midcurve-api/src/app/api/v1/auth/verify/route.ts:140
let user = await getAuthUserService().findUserByWallet(address, chainId);
```

This treats the same address on different chains as different wallets, leading to duplicate user creation.

## Why This Is Wrong

The same private key controls an address across **ALL EVM chains**. The `chainId` in SIWE just indicates which network the user's wallet was connected to when signing - it doesn't change ownership.

If someone can sign a SIWE message for `0xABC...` on Arbitrum, they own that address on Ethereum, Base, Polygon, and all other EVM chains.

## Proposed Solution

Remove `chainId` from `AuthWalletAddress` entirely, making `address` the sole unique identifier.

### Files Affected

- `packages/midcurve-database/prisma/schema.prisma` - Remove chainId, update unique constraint
- `packages/midcurve-shared/src/types/auth-wallet-address.ts` - Remove chainId from type
- `packages/midcurve-api-shared/src/types/auth/signup.ts` - Remove chainId from request/response
- `packages/midcurve-api-shared/src/types/auth/link-wallet.ts` - Remove chainId from LinkWalletData
- `packages/midcurve-services/src/services/auth/auth-user-service.ts` - Remove chainId from methods
[...]

PR #68, opened 2026-05-05
feat: add UniswapV3StakingPositionService (SPEC-0003b PR4a)

0xNedAlbo:
## Summary

PR4a of 5 for SPEC-0003b. Adds the service layer for `UniswapV3StakingVault` positions; mirrors `UniswapV3VaultPositionService` (closer-in-spirit canonical) and reuses PR2's `syncFromChain` for chain ingestion + domain-event publishing.

PR4b (REST endpoints + serializers + cross-protocol wire-up) will land separately.

## Refinements applied (per maintainer feedback on the PR4 plan)

1. **No 15-second `refresh()` cache.** `refresh()` always does a fresh chain pull. The 60s block-keyed RPC cache stays inside `refreshOnChainState` — it's keyed by block number so it can't serve stale data for the same block.
2. **`reset` drops revert event emission.** FK cascade is the documented cleanup path; PR3's rule's revert handler is a no-op; the immediate `syncFromChain` call re-emits liquidity events that rebuild lots and journal entries from scratch. Inline code comment justifies the divergence from NFT/Vault.
3. **`reset` test boundary assertion** added — verifies the publisher call set after `reset` matches the original import (ledger service called exactly once for rebuild + zero revert events).
4. **`discoverWalletPositions` `fromBlock`** uses `0n` (matches the Vault precedent + PR2's `findVaultBirthBlock` pattern). Per-chain deploy block can be added to `SharedContract.config` in a follow-up if RPC limits become a concern.

### `packages/midcurve-services/src/services/position/uniswapv3-staking-position-service.ts` (new, ~870 LoC)

Constructor DI: optional `prisma`, `eventPublisher`, `evmConfig`, `poolService`, `evmBlockService`, `poolPriceService`, `cacheService`, `sharedContractService`, `erc20TokenService`. Singletons for config/publisher; new instances for sub-services. `UniswapV3StakingLedgerService` is instantiated per-call (matches existing pattern).

Public surface:
- `discover(userId, { chainId, vaultAddress, quoteTokenAddress? })` — single-position importer. Owner derived from `vault.owner()` on chain. Throws `INVALID_VAULT_CONTRACT` on chain-read failure. Calls `syncFromChain` via `refresh()` to import history + emit domain events.
- `discoverWalletPositions(userId, walletAddress, chainIds?)` — factory scan. `VaultCreated(owner=walletAddress)` filter, `fromBlock: 0n`. Chains with no registered factory → `{ found: 0 }` gracefully.
- `refresh(id, blockNumber? = 'latest')` — always fresh chain pull. **No `updatedAt` short-circuit.**
- `reset(id)` — wipes ledger via `ledgerService.deleteAll()` then reimports via `syncFromChain`. No revert events emitted.
- `refreshOnChainState` — block-keyed 60s RPC cache. `Staked → Settled` transition emits `position.closed`.
- `delete(id)` — emits `position.deleted` in a transaction.
- `fetchMetrics(id)` — non-persisted snapshot returning `UniswapV3StakingPositionMetrics`.
- `findById`, `findByPositionHash` — standard CRUD.

### `packages/midcurve-shared/src/types/position/uniswapv3-staking/uniswapv3-staking-position-metrics.ts` (new)

`UniswapV3StakingPositionMetrics` — parallel to `UniswapV3VaultPositionMetrics` with staking-specific fields (`vaultState`, `yieldTarget`, `pendingBps`).

## PR4a limitation

`unclaimedYieldBase/Quote` in state is set to `0n`. The contract's `_expectedBalancesAfterPartialClose(uint16)` is internal, so we cannot compute the live yield split without a public view. A SPEC-0003a follow-up can expose `quoteCloseAt(uint16)` or similar; until then, `0n` is a safe default and the rest of the state stays accurate.

## Tests (14)

- `discover` happy path → creates Position row, calls `syncFromChain`, returns position. **Asserts `position.created` is NOT emitted from the service** (matches NFT/Vault convention — emitted by the API route in PR4b).
- `discover` chain reads fail → throws `INVALID_VAULT_CONTRACT`.
- `discoverWalletPositions` chain with no factory → `{ found: 0 }` gracefully.
- `discoverWalletPositions` `getLogs` filtered by `owner=walletAddress` with `fromBlock: 0n`.
- `discoverWalletPositions` matching `VaultCreated` log triggers `discover()`.
- `refresh` always calls `syncFromChain` + `refreshOnChainState` (no `updatedAt` short-circuit) — refinement #1.
- `refreshOnChainState` `Staked → Settled` emits `position.closed` once.
- `refreshOnChainState` no transition → no close event.
- `refreshOnChainState` block-keyed cache hit → no extra readContract calls.
- `reset` wipes ledger, **does NOT emit `position.liquidity.reverted`** — refinement #3.
- `reset` re-publishes the same liquidity events as the original import (boundary assertion at the ledger-service / publisher).
- `delete` emits `position.deleted` in a transaction.
- `fetchMetrics` returns vault-specific snapshot.
- `findByPositionHash` uses positionHash + protocol filter.

Mocks: `vi.hoisted` prismaMock, module mocks for `@midcurve/database`, `../../events`, and the staking ledger service. No live RPC, no live RabbitMQ, no live DB.

## Out of scope (deferred to PR4b)

- REST endpoints under `/positions/uniswapv3-staking/...`
- Wire types + serializer in `@midcurve/api-shared`
- Service factory in `apps/midcurve-api/src/lib/services.ts`
- Cross-protocol updates: extend top-level `/positions/discover` for wallet scan, extend `/positions/refresh-all` for staking dispatch
- List-view smoke test for `state.yieldTarget` / `state.vaultState` pass-through

Refs #63

## Test plan
- [x] `pnpm typecheck` — all 14 packages pass
- [x] `pnpm --filter @midcurve/services test:run` — 178 tests pass (14 new staking position service tests)
- [x] `pnpm --filter @midcurve/business-logic test:run` — 12 tests pass (no regressions)
- [x] `pnpm --filter @midcurve/shared test:run` — 127 tests pass (no regressions)

### DIFF
--- a/packages/midcurve-services/src/services/position/index.ts
+++ b/packages/midcurve-services/src/services/position/index.ts
@@ -15,5 +15,8 @@ export type {
 export { UniswapV3VaultPositionService } from './uniswapv3-vault-position-service.js';
 export type { UniswapV3VaultPositionServiceDependencies } from './uniswapv3-vault-position-service.js';
 
+export { UniswapV3StakingPositionService } from './uniswapv3-staking-position-service.js';
+export type { UniswapV3StakingPositionServiceDependencies } from './uniswapv3-staking-position-service.js';
+
 export { PositionArchiveService } from './position-archive-service.js';
 export type { PositionArchiveServiceDependencies } from './position-archive-service.js';
--- a/packages/midcurve-services/src/services/position/uniswapv3-staking-position-service.test.ts
+++ b/packages/midcurve-services/src/services/position/uniswapv3-staking-position-service.test.ts
@@ -490,20 +490,23 @@ describe('UniswapV3StakingPositionService', () => {
     });
 
     // -------------------------------------------------------------------------
-    // 6. refresh — always orchestrates ledger sync + state refresh (NO updatedAt cache)
+    // 6. refresh — no updatedAt short-circuit (would catch a regression that
+    //    re-introduces the 15-second cache).
     // -------------------------------------------------------------------------
-    it('refresh — always calls syncFromChain + refreshOnChainState (no updatedAt short-circuit)', async () => {
-        // Position was just updated 1ms ago — would be a cache hit if we had a 15s cache.
+    it('refresh — two back-to-back calls both invoke syncFromChain (no updatedAt cache)', async () => {
+        // Position was updated 1ms ago — a hypothetical 15s cache would short-circuit.
         const recentRow = makePositionRow({ updatedAt: new Date(Date.now() - 1) });
         prismaMock.position.findFirst.mockResolvedValue(recentRow);
 
         const { service } = makeService({});
         await service.refresh(POSITION_ID);
+        await service.refresh(POSITION_ID);
 
-        // refreshAllPositionLogs called → syncFromChain called
-        expect(mockLedgerInstance.syncFromChain).toHaveBeenCalledTimes(1);
-        // refreshOnChainState ran → position.update called
-        expect(prismaMock.position.update).toHaveBeenCalled();
+        // The invariant: refresh always pulls fresh chain state. A regression
+        // that re-introduces the 15s cache would yield only 1 call here.
+        expect(mockLedgerInstance.syncFromChain).toHaveBeenCalledTimes(2);
+        // refreshOnChainState ran on each call → position.update called twice.
+        expect(prismaMock.position.update).toHaveBeenCalledTimes(2);
     });
 
     // -------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


0xNedAlbo:
Solid PR overall — refinements 1, 3, and 5 are correctly implemented with clear inline justification. Six points before merge:

**1. ⚠️ `discover` must drop the `quoteTokenAddress` parameter entirely.** `isToken0Quote` is on-chain immutable for staking vaults — encoded at `stake()` time and never mutable afterward. The optional `quoteTokenAddress` override on `discover()` is not just unnecessary, it's a latent correctness bug: PR1's ledger service uses `isToken0Quote` as ground truth in every disposal calculation (`valueOfBaseAndQuote`), so a caller passing a `quoteTokenAddress` that disagrees with `vault.isToken0Quote()` silently corrupts all subsequent PnL math. The "parity with the vault service signature" justification doesn't apply — vault shares aren't governed by an on-chain immutable here, staking is. Remove the parameter from the signature; the quote token always comes from the chain via `vault.isToken0Quote()`.

**2. Test #11 boundary is weaker than the plan claimed.** The plan said "publisher call log shows the same set of `position.liquidity.{increased,decreased}` events (same `ledgerInputHash` set, same count, same chain order) as the original import". The actual test only asserts `syncFromChain.calls.length === initialSyncCalls + 1`. Because `syncFromChain` is mocked, the test cannot verify the publisher call set — that's a real limitation of unit-testing across the package boundary. Two options: (a) accept the limitation explicitly in a test comment ("re-emission correctness is verified at PR2's syncFromChain test layer"), or (b) un-mock `syncFromChain` for this one test and assert the publisher calls. Currently the test reads as stronger than it is.

**3. Test #6 doesn't actually verify the no-cache invariant.** Setting `updatedAt: Date.now() - 1` and asserting `syncFromChain` is called once is consistent with the no-cache implementation, but it's also consistent with a hypothetical 15s cache that just hasn't been hit yet (one call passes either way). The test that would catch a regression: call `refresh()` twice in quick succession and assert `syncFromChain.calls.length === 2`. Worth tightening.

**4. `recalculateAggregates` runs ~4x per `refresh()` call.** Production path: `refresh` → `refreshAllPositionLogs` → `syncFromChain` → `importLogsForPosition` (which calls `recalculateAggregates` twice, pre + post) → back to `refresh` → `refreshOnChainState` (calls `recalculateAggregates`) → optionally `fetchMetrics` (calls it again). For a settled vault with 50+ swap events that's a lot of redundant O(n) DB iteration with n×UPDATE writes per pass. PR4b can address this with caller-side memoization, but worth flagging now so the design doesn't ossify. Could `refresh` pass the post-import aggregates from `syncFromChain` down to `refreshOnChainState` to skip one full pass?

**5. `unclaimedYield = 0n` needs a tracking issue.** The PR honestly notes that `_expectedBalancesAfterPartialClose` is internal so unclaimed-yield split can't be computed. Without a follow-up to expose `quoteCloseAt(uint16 bps)` (or similar) on the contract, the UI will show "0 unclaimed yield" for every active staking position — strictly worse than the equivalent NFT view. Please open a SPEC-0003a follow-up issue and link it from this PR, so the gap doesn't get forgotten.

**6. RPC scan limits — confirmed follow-up: persist `deployBlock` in `SharedContract.config`.** `findVaultBirthBlock` and `discoverWalletPositions` both do `fromBlock: 0n → toBlock: 'latest'` scans. Fine on freshly-deployed Arbitrum factory; at scale this hits RPC limits. Confirmed follow-up: extend `SharedContract.config` with a `deployBlock: number` field (set at registration time) and use it as the default `fromBlock` for both scan paths. Please open a tracking issue and link it from this PR.

Points 1 and 2 should land before merge. 3 is a nice-to-have. 4 is a design discussion that can wait but shouldn't be silently inherited by PR4b. 5 and 6 are tracking issues to open.

0xNedAlbo:
Pushed `a93ae193` addressing the review points.

**Code changes (must-land):**

1. **Dropped `quoteTokenAddress` from `discover()`** — agreed it was a latent correctness bug. `isToken0Quote` now comes exclusively from `vault.isToken0Quote()` on chain. Inline doc comment justifies the divergence from `UniswapV3VaultPositionService.discover`. (Point #1)
2. **Tightened test #6** — now calls `refresh()` twice in succession and asserts `syncFromChain.calls.length === 2`. A regression that re-introduces a 15s cache would yield 1 here. (Point #3)
3. **Acknowledged test #11 boundary limitation** — added an explicit comment that `syncFromChain` is mocked, so the planned `ledgerInputHash`-set comparison isn't possible across the package boundary. Re-emission correctness is verified at PR2's `syncFromChain` test layer; PR3's rule consumes those events deterministically via `journalService.isProcessed`, so a correct re-emission from PR2 + FK cascade from `deleteAll` yields a correct journal rebuild end-to-end. The test now asserts what's actually verifiable here. (Point #2)

**Tracking issues opened:**

- #69 — expose `quoteCloseAt(uint16)` on `UniswapV3StakingVault` so the position service can populate `unclaimedYieldBase/Quote` instead of returning `0n`. Includes the proposed Solidity signature and the off-chain consumer changes that will land once the contract is amended. (Point #5)
- #70 — persist `deployBlock` in `SharedContract.config` to bound the `fromBlock: 0n → 'latest'` factory scans in `findVaultBirthBlock` (PR2) and `discoverWalletPositions` (PR4a, Vault). Lists the call sites that will pick up the bound. (Point #6)

**Design discussion deferred to PR4b (Point #4):**

`recalculateAggregates` currently runs ~4x per `refresh()` call:
- `refresh` → `refreshAllPositionLogs` → `syncFromChain` → `importLogsForPosition` (pre + post recalc)
- Then `refresh` → `refreshOnChainState` (recalc again)
- Plus `fetchMetrics` (recalc again)

Cleanest fix: `syncFromChain` already returns `postImportAggregates` in `StakingImportLogsResult`. Threading that through from `refreshAllPositionLogs` → `refreshOnChainState` would skip one full O(n) replay. PR4b's REST endpoints don't need to change shape; this is purely an internal optimization. I'll include it in PR4b and add a focused test (`recalculateAggregates is called at most twice across a full refresh`).

All 14 tests still green; 14 packages typecheck. Ready for re-review.

0xNedAlbo:
All three merge-blocker points landed cleanly:

**1. `quoteTokenAddress` removed from `discover` ✓** Signature is now `discover(userId, { chainId, vaultAddress })` and the doc-comment explicitly explains the staking-specific divergence ("on-chain immutable... a caller-supplied override that disagrees with `vault.isToken0Quote()` would silently corrupt all subsequent PnL math"). Exactly the right framing.

**2. Test #6 properly verifies the no-cache invariant ✓** Two back-to-back `refresh()` calls with `syncFromChain.calls.length === 2` and `position.update.calls.length === 2`. A regression that re-introduces the 15s cache would now fail this test loudly.

**3. Test #11 honest about the boundary limitation ✓** The new test name (`invokes deleteAll + syncFromChain exactly once each, no revert events`) and the prefixed ⚠️ block in the test comment make it clear what the unit test can and cannot verify, and where re-emission correctness actually gets tested (PR2's syncFromChain test layer + PR3's idempotency via `journalService.isProcessed`). Good.

Points 4 (recalculate-4x), 5 (`unclaimedYield = 0n`), and 6 (`deployBlock` follow-up) remain open as noted — none of them block the merge.

LGTM. Ship it.

PR #78, opened 2026-08-08
fix: extend close-order processing to uniswapv3-vault positions

### LINKED ISSUE #77: Vault close orders never reach the database — extend close-order processing to uniswapv3-vault
## Context

Stop-loss and take-profit orders work for `uniswapv3` (NFT) positions. For `uniswapv3-vault` (tokenized) positions the feature is complete on-chain, in the UI and in the executor, but the step that writes the order into the database is NFT-only. A vault order therefore registers successfully on-chain and then disappears: no `CloseOrder` row is created, the monitor has nothing to watch, and the already-finished vault execution path never runs. The user sees a successful transaction followed by an empty order list.

## Where the chain breaks

`UniswapV3ProcessCloseOrderEventsRule` binds both routing patterns — `closer.*.*.*` and `closer.vault.*.*.*` — and its documentation claims to handle both variants. Every handler underneath is NFT-only:

| Location | Problem |
|---|---|
| `resolveOrder`, `handleRegistered` | throw when `event.nftId` is absent, which is always the case for vault events |
| `createOrderIdentityHash` | hard-codes the `uniswapv3/{chainId}/{nftId}/{triggerMode}` shape |
| position lookup | filters on `protocol: 'uniswapv3'` and `config.nftId` |
| `buildCreateInput` | writes `protocol: 'uniswapv3'` |
| injected position service | `UniswapV3PositionService`, used for the order tag |

The thrown error is caught in `handleMessage` and the message is nacked without requeue, so it goes straight to the dead-letter exchange. Nothing surfaces the loss.

Everything upstream already works: `close-order-receipt-publisher.ts` recognises vault events and routes them as `closer.vault.{chainId}.{vaultAddress}.{triggerMode}`, and `UniswapV3VaultPositionCloser` is registered as a shared contract name.

## Decision: one identity scheme

Vault orders use the same identity scheme as NFT orders — this was the intent from the start. An order identity is the pair *(position, triggerMode)*; the hash only encodes what determines the position.

One consequence must be respected: `(chainId, vaultAddress, triggerMode)` is **not** unique. The vault is an ERC-20 whose shares can be held by several owners, each able to hold their own orders — the executor reads `getOrder(vault, owner, triggerMode)` and derives liquidity from `balanceOf(owner)` against `totalSupply`. The position discriminator for a vault therefore comprises vault address **and** owner address, matching the vault position hash `uniswapv3-vault/{chainId}/{vaultAddress}/{ownerAddress}`.

The routing key may stay as it is. It distributes messages; it does not identify anything. All events land in the same queue and the owner travels in the payload.

The two-slots-per-position invariant is unchanged: exactly one LOWER and one UPPER slot per position, and re-registration on an occupied slot replaces the previous order.

## Required behaviour

| Aspect | Requirement |
|---|---|
| Order identity | built from protocol, chain, position discriminator and trigger mode; for vault the discriminator includes the owner |
| Position resolution | a vault event must resolve to exactly one position with `protocol: 'uniswapv3-vault'`, matched on chain, vault address and owner |
| `protocol` on the order | must be `'uniswapv3-vault'`, otherwise the executor cannot dispatch to its existing vault path |
| Order tag for automation logs | must use the vault position's `isToken0Quote` and token decimals |
| Unroutable message | a message no handler can process must be observably distinguishable from a processed one |

The semantic `closeOrderHash` (`sl@tick` / `tp@tick`) is already protocol-independent and stays as it is. Existing NFT order rows keep their current identity hashes; no data migration is expected.

## Secondary gaps in the same package

The fallback poller does not cover vault orders: `CloseOrderSubscriber` loads only contracts named `UniswapV3PositionCloser`, and its poll loop skips every event without an `nftId`. The safety net for direct contract interaction must cover the vault closer too.

The read API is missing the single-order endpoint for vault positions. The NFT variant has `close-orders/{closeOrderHash}`; the vault variant only has the list endpoint and the nested `automation-state` route.

## To verify before implementing

Whether the close-order event decoder populates the owner in the payload for vault registration events. The NFT path writes `payload.owner` into the order state; without it, position resolution is impossible and the decoder needs extending first. Report findings before changing behaviour.

## Out of scope

- The accounting layer (journal, token lots, wallet perimeter) — separate work, no dependency in either direction
- Changes to the on-chain contracts
- Changes to the NFT order path beyond what sharing the scheme requires
- UI changes — the vault risk-triggers wizard and order panel already exist

## Acceptance criteria

1. Registering a stop-loss on a vault position creates exactly one `CloseOrder` row with `protocol: 'uniswapv3-vault'`, linked to the correct vault position.
2. Two owners holding shares in the same vault can each hold their own LOWER and UPPER orders without colliding.
3. Re-registering on an occupied slot replaces the previous order rather than creating a second one.
4. Cancelling a vault order on-chain removes the row and records a cancellation in the automation log.
5. Executing a vault order removes the row, preserves execution data in the automation log, and the monitor's subscription for that order is removed.
6. Trigger-tick, slippage, operator, payout, valid-until and swap-intent updates on a vault order are reflected in the stored order.
7. Replaying the same event twice produces no duplicate rows and no duplicate log entries.
8. The fallback poller picks up a vault order registered by direct contract interaction, without the UI involved.
9. `GET` on a single vault close order by its `closeOrderHash` returns the same shape as the NFT equivalent.
10. A message that no handler can process is visible as such — not silently indistinguishable from success.

0xNedAlbo:
Closes #77

Vault close orders now reach the database. Implemented per the decisions in [this comment](https://github.com/0xNedAlbo/midcurve-finance/issues/77#issuecomment-5225632075).

## What was broken

`processEvent` switched on the full event type string and matched only `close-order.*.uniswapv3`. Every `…-vault` event fell through to `default` → `warn` → **`ack`**: acknowledged and discarded, never reaching `resolveOrder` or `handleRegistered`. The queue was also declared without `x-dead-letter-exchange`, so the `nack` path in the catch block dropped messages as well. Nothing surfaced either loss.

## Commits

| Commit | Scope |
|---|---|
| `8bc5f7bb` | `@midcurve/services` — shared identity + routing helpers, `upsertByIdentityHash`, shares typing, close-order DLX |
| `0fab181e` | business-logic — the actual fix, plus tests |
| `c4ae2bc6` | onchain-data — fallback poller and catch-up cover the vault closer |
| `ac9bd6a7` | api — single vault close-order endpoint |
| `34de883b` | review: restore per-event isolation in the poll loop |
| `9941038d` | review: keep `sharesClosed` in the vault execution log |
| `c1b5b0d4` | review: correct the v1 queue note — it fills, it does not drain |

## Decisions applied

- **Q1 — EIP-55.** `createCloseOrderIdentityHash()` is the one place that decides the hash shape and normalizes addresses; `refresh()` now calls it instead of lowercasing its own template literal. NFT hashes contain no address and are untouched.
- **Q2 — disposition split.** *Understood, doesn't concern us* (no matching position, no matching order) → `info` + ack. *Cannot be interpreted* (unknown type, missing identifiers, ambiguous position) → `error` + DLQ. An untracked share holder is the normal case for a vault, so it does not log at `warn`.
- **Q4 — `OrderSharesUpdated` handled**, merging `state.shares` (string, per the bigint rule) and writing an `ORDER_MODIFIED` log entry.
- **Q5 — `OrderExecuted` added to the poller's topic filter.** It was missing for *both* protocols, so the fallback path had never replayed an execution.
- **Q6 — new queue name.** `…-process-close-order-events.v2`, with `x-dead-letter-exchange` → `close-order-events-dlx` → `…-process-close-order-events.dlq` (7-day TTL). The v1 queue must be deleted at deploy time — it fills rather than drains, see deploy steps.
- **Q7 — asymmetry kept.** The rule still does not set `sharedContractId`; NFT behaviour unchanged. Filed as follow-up.
- **Q8 — new endpoint normalizes its path addresses.** Siblings unchanged; filed as follow-up.
- **R2 — registration is an upsert.** `upsertByIdentityHash()` states the operation as "this slot now holds this order", so whichever of the two writers wins the race, the slot ends up with the same content. No `try/catch`, no P2002 handling.
- **R4 — out of scope.** No dedupe key on the automation log. Filed as follow-up.

Also injected the rule's collaborators (defaults unchanged in production) — constructing `UniswapV3PositionService` calls `EvmConfig.getInstance()`, which throws without `initAppConfig()`, so the rule was untestable otherwise.

## Verification

- `pnpm typecheck` — 14/14 tasks pass
- `pnpm build` — 11/11 pass; the new route appears in the API route table
- business-logic: 36 tests pass (24 new)
- services: 199 tests pass (10 new on the identity helper)

The new rule tests map onto the acceptance criteria: one row with `protocol: 'uniswapv3-vault'` linked to the right position (1); two owners on one vault get separate slots, and LOWER/UPPER stay separate (2); re-registration replaces (3); cancel and execute delete the row, log, and drop the subscription (4, 5); all seven update events land in `state` (6); a replayed registration creates nothing (7); an uninterpretable message nacks while an understood-but-not-ours message acks (10). Plus: `sharesClosed` reaches the execution log context for a vault order and is omitted entirely for an NFT one, the order tag is built from the vault position service rather than the NFT one, and the NFT identity hash is pinned unchanged.

**AC 8 is not unit-tested** — the fallback poller needs a chain. Manual verification steps are below; I have not run them yet.

## Deploy steps

Both are required. Neither is hygiene.

### 1. Delete legacy vault close-order rows — before or with the deploy, never after

```sql
DELETE FROM close_orders WHERE protocol = 'uniswapv3-vault';
```

This is load-bearing. `upsertByIdentityHash` keys on `orderIdentityHash`, but the collision risk sits on the *other* constraint, `@@unique([positionId, closeOrderHash])`. A legacy lowercase row is invisible to the upsert's key, so the upsert inserts a second row for the same slot and then dies on `positionId + closeOrderHash` — dead-lettered, and dead-lettered again on every replay. The upsert protects against the race between the two writers; it does not protect against the old data. Rows deleted here are re-derived by `refresh()`, which is why deleting them is safe.

Scoped to `protocol = 'uniswapv3-vault'` on purpose — NFT identity hashes contain no address and must not be touched. The dev database holds zero close-order rows of either protocol.

### 2. Delete the v1 queue

```
rabbitmqadmin delete queue name=business-logic.uniswapv3-process-close-order-events
```

The v1 queue does **not** drain. It is durable, `autoDelete: false`, and still bound to `close-order-events` with both routing patterns; a topic exchange delivers to every matching binding, so from the moment the service moves to `.v2` the old queue receives every close-order event with nobody consuming it, and grows until the disk does. Do this at deploy time, not "eventually".

Not done from code on purpose: deleting a queue that may still hold unprocessed v1 messages is destructive, and mid-rolling-deploy is the worst moment to decide that. Say the word if you would rather the rule unbind it at startup.

## Manual verification still owed (AC 8 + Q3)

On Arbitrum against `UniswapV3VaultPositionCloser` `0x13d13B15BbE9b06C0279a7aB5f0a898EA3f25A40`:

1. Register a stop-loss on a vault position through the wizard → confirm exactly one `close_orders` row with `protocol='uniswapv3-vault'` and an EIP-55 identity hash.
2. Register directly against the contract, without the UI, and wait for a poll cycle → confirm the row appears (AC 8).
3. While doing so, record whether the reconciliation path (`UniswapV3VaultPositionService.refresh()` → `closeOrderService.refresh()`) also produces the row. Its `catch` downgrades any failure to a `warn`, and per Q3 we left that behaviour alone — this settles empirically whether a second bug hides there.
4. Confirm the dead-letter path end to end by publishing an uninterpretable message to `close-order-events` and checking it lands in `…-process-close-order-events.dlq`. `onStartup` is the one part of the DLX wiring no test covers.

I will post the results on the issue.

### DIFF
--- a/apps/midcurve-api/src/app/api/v1/positions/uniswapv3-vault/[chainId]/[vaultAddress]/[ownerAddress]/close-orders/[closeOrderHash]/route.ts
+++ b/apps/midcurve-api/src/app/api/v1/positions/uniswapv3-vault/[chainId]/[vaultAddress]/[ownerAddress]/close-orders/[closeOrderHash]/route.ts
@@ -1,0 +1,194 @@
+/**
+ * Single Vault Close Order Endpoint (by semantic identifier)
+ *
+ * GET /api/v1/positions/uniswapv3-vault/:chainId/:vaultAddress/:ownerAddress/close-orders/:closeOrderHash
+ *
+ * Vault counterpart of the NFT route
+ * /api/v1/positions/uniswapv3/:chainId/:nftId/close-orders/:closeOrderHash —
+ * same response shape, same error codes.
+ *
+ * Authentication: Required (session only)
+ *
+ * Note: no PUT/PATCH/DELETE — the UI calls the contract directly via Wagmi, and
+ * the event subscriber (UniswapV3ProcessCloseOrderEventsRule) handles all DB writes.
+ */
+
+import { NextRequest, NextResponse } from 'next/server';
+import { z } from 'zod';
+import { withAuth } from '@/middleware/with-auth';
+import {
+  createSuccessResponse,
+  createErrorResponse,
+  ApiErrorCode,
+  ErrorCodeToHttpStatus,
+  CloseOrderHashSchema,
+} from '@midcurve/api-shared';
+import { normalizeAddress } from '@midcurve/shared';
+import { serializeCloseOrder } from '@/lib/serializers';
+import { apiLogger, apiLog } from '@/lib/logger';
+import {
+  getUniswapV3CloseOrderService,
+  getUniswapV3VaultPositionService,
+} from '@/lib/services';
+import { createPreflightResponse } from '@/lib/cors';
+
+export const runtime = 'nodejs';
+export const dynamic = 'force-dynamic';
+
+/**
+ * Path params schema.
+ *
+ * Addresses are normalized to EIP-55 so a lowercase address in the URL resolves
+ * the same position as a checksummed one — the position hash is built from
+ * checksummed addresses.
+ */
+const PathParamsSchema = z.object({
+  chainId: z.string().regex(/^\d+$/).transform(Number),
+  vaultAddress: z
+    .string()
+    .regex(/^0x[0-9a-fA-F]{40}$/)
+    .transform((value) => normalizeAddress(value)),
+  ownerAddress: z
+    .string()
+    .regex(/^0x[0-9a-fA-F]{40}$/)
+    .transform((value) => normalizeAddress(value)),
+  closeOrderHash: CloseOrderHashSchema,
+});
+
+/**
+ * Handle CORS preflight
+ */
+export async function OPTIONS(request: NextRequest): Promise<Response> {
+  return createPreflightResponse(request.headers.get('origin'));
+}
+
+/**
+ * GET /api/v1/positions/uniswapv3-vault/:chainId/:vaultAddress/:ownerAddress/close-orders/:closeOrderHash
+ *
+ * Get a specific vault close order by its semantic identifier.
+ *
+ * Path parameters:
+ * - chainId: EVM chain ID
+ * - vaultAddress: Vault (ERC-20 share token) address
+ * - ownerAddress: Share holder address
+ * - closeOrderHash: Semantic identifier (e.g., "sl@-12345", "tp@201120")
+ */
+export async function GET(
+  request: NextRequest,
+  {
+    params,
+  }: {
+    params: Promise<{
+      chainId: string;
+      vaultAddress: string;
+      ownerAddress: string;
+      closeOrderHash: string;
+    }>;
+  }
+): Promise<Response> {
+  return withAuth(request, async (user, requestId) => {
+    const startTime = Date.now();
+
+    try {
+      // 1. Parse and validate path parameters
+      const resolvedParams = await params;
+      const paramsValidation = PathParamsSchema.safeParse(resolvedParams);
+
+      if (!paramsValidation.success) {
+        apiLog.validationError(apiLogger, requestId, paramsValidation.error.errors);
+
+        const errorResponse = createErrorResponse(
+          ApiErrorCode.VALIDATION_ERROR,
+          'Invalid path parameters',
+          paramsValidation.error.errors
+        );
+
+        apiLog.requestEnd(apiLogger, requestId, 400, Date.now() - startTime);
+
+        return NextResponse.json(errorResponse, {
+          status: ErrorCodeToHttpStatus[ApiErrorCode.VALIDATION_ERROR],
+        });
+      }
+
+      const { chainId, vaultAddress, ownerAddress, closeOrderHash } =
+        paramsValidation.data;
+
+      // 2. Find vault position by positionHash
+      const positionHash = `uniswapv3-vault/${chainId}/${vaultAddress}/${ownerAddress}`;
+      const position = await getUniswapV3VaultPositionService().findByPositionHash(
+        user.id,
+        positionHash
+      );
+
+      if (!position) {
+        const errorResponse = createErrorResponse(
+          ApiErrorCode.POSITION_NOT_FOUND,
+          'Vault position not found',
+          `No vault position found for chainId ${chainId}, vaultAddress ${vaultAddress} and ownerAddress ${ownerAddress}`
+        );
+
+        apiLog.requestEnd(apiLogger, requestId, 404, Date.now() - startTime);
+
+        return NextResponse.json(errorResponse, {
+          status: ErrorCodeToHttpStatus[ApiErrorCode.POSITION_NOT_FOUND],
+        });
+      }
+
+      apiLog.businessOperation(
+        apiLogger,
+        requestId,
+        'get',
+        'close-order',
+        closeOrderHash,
+        { chainId, vaultAddress, ownerAddress, positionId: position.id }
+      );
+
+      // 3. Find close order by position + hash
+      const order = await getUniswapV3CloseOrderService().findByPositionAndHash(
+        position.id,
+        closeOrderHash
+      );
+
+      if (!order) {
+        const errorResponse = createErrorResponse(
+          ApiErrorCode.NOT_FOUND,
+          'Close order not found',
+          `No close order found with hash ${closeOrderHash}`
+        );
+
+        apiLog.requestEnd(apiLogger, requestId, 404, Date.now() - startTime);
+
+        return NextResponse.json(errorResponse, {
+          status: ErrorCodeToHttpStatus[ApiErrorCode.NOT_FOUND],
+        });
+      }
+
+      // 4. Serialize and return
+      const serialized = serializeCloseOrder(order);
+      const response = createSuccessResponse(serialized);
+
+      apiLog.requestEnd(apiLogger, requestId, 200, Date.now() - startTime);
+
+      return NextResponse.json(response, { status: 200 });
+    } catch (error) {
+      apiLog.methodError(
+        apiLogger,
+        'GET /api/v1/positions/uniswapv3-vault/:chainId/:vaultAddress/:ownerAddress/close-orders/:closeOrderHash',
+        error,
+        { requestId }
+      );
+
+      const errorResponse = createErrorResponse(
+        ApiErrorCode.INTERNAL_SERVER_ERROR,
+        'Failed to get close order',
+        error instanceof Error ? error.message : String(error)
+      );
+
+      apiLog.requestEnd(apiLogger, requestId, 500, Date.now() - startTime);
+
+      return NextResponse.json(errorResponse, {
+        status: ErrorCodeToHttpStatus[ApiErrorCode.INTERNAL_SERVER_ERROR],
+      });
+    }
+  });
+}
--- a/apps/midcurve-business-logic/src/rules/close-orders/uniswapv3/close-order-event-types.ts
+++ b/apps/midcurve-business-logic/src/rules/close-orders/uniswapv3/close-order-event-types.ts
@@ -28,5 +28,17 @@ export type {
   OrderValidUntilUpdatedEvent,
   OrderSlippageUpdatedEvent,
   OrderSwapIntentUpdatedEvent,
+  OrderSharesUpdatedPayload,
+  // Vault variants — same payloads, vault/owner identifiers in the envelope
+  VaultOrderRegisteredEvent,
+  VaultOrderCancelledEvent,
+  VaultOrderExecutedEvent,
+  VaultOrderOperatorUpdatedEvent,
+  VaultOrderPayoutUpdatedEvent,
+  VaultOrderTriggerTickUpdatedEvent,
+  VaultOrderValidUntilUpdatedEvent,
+  VaultOrderSlippageUpdatedEvent,
+  VaultOrderSwapIntentUpdatedEvent,
+  VaultOrderSharesUpdatedEvent,
   AnyCloseOrderEvent,
 } from '@midcurve/services';
--- a/apps/midcurve-business-logic/src/rules/close-orders/uniswapv3/uniswapv3-process-close-order-events.test.ts
+++ b/apps/midcurve-business-logic/src/rules/close-orders/uniswapv3/uniswapv3-process-close-order-events.test.ts
@@ -1,0 +1,638 @@
+/**
+ * UniswapV3ProcessCloseOrderEventsRule — unit tests
+ *
+ * Issue #77 acceptance criteria, vault variant:
+ *  1. Registering a stop-loss on a vault position creates exactly one row with
+ *     protocol 'uniswapv3-vault', linked to the correct vault position
+ *  2. Two owners on the same vault hold their own LOWER/UPPER slots
+ *  3. Re-registration on an occupied slot replaces the previous order
+ *  4. Cancellation removes the row and logs the cancellation
+ *  5. Execution removes the row, preserves execution data, drops the subscription
+ *  6. Trigger-tick, slippage, operator, payout, valid-until, swap-intent and
+ *     shares updates land in the stored order
+ *  7. Replaying the same event twice produces no duplicate rows
+ * 10. A message no handler can process is dead-lettered, not acked
+ *
+ * No live RabbitMQ, no live Prisma. The database module is mocked and the rule's
+ * own service instances are replaced after construction.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// =============================================================================
+// MODULE MOCKS
+// =============================================================================
+
+const { prismaMock, txMock } = vi.hoisted(() => {
+  const txMock = {
+    position: {
+      findUnique: vi.fn<(args: unknown) => Promise<unknown>>(),
+      findMany: vi.fn<(args: { where: Record<string, unknown> }) => Promise<Array<{ id: string; config: Record<string, unknown> }>>>(),
+    },
+  };
+  return {
+    txMock,
+    prismaMock: {
+      position: { findUnique: vi.fn(), findMany: vi.fn() },
+      $transaction: vi.fn(async (cb: (tx: unknown) => unknown) => cb(txMock)),
+    },
+  };
+});
+vi.mock('@midcurve/database', () => ({
+  prisma: prismaMock,
+}));
+
+import { ContractTriggerMode, ContractSwapDirection } from '@midcurve/shared';
+import { getDomainEventPublisher } from '@midcurve/services';
+import { UniswapV3ProcessCloseOrderEventsRule } from './uniswapv3-process-close-order-events';
+
+// =============================================================================
+// FIXTURES
+// =============================================================================
+
+const CHAIN_ID = 42161;
+const VAULT = '0x13d13B15BbE9b06C0279a7aB5f0a898EA3f25A40';
+const OWNER_A = '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48';
+const OWNER_B = '0xaf88d065e77c8cC2239327C5EDb3A432268e5831';
+const POOL = '0xC6962004f452bE9203591991D15f6b388e09E8D0';
+const CLOSER = '0x13d13B15BbE9b06C0279a7aB5f0a898EA3f25A40';
+const OPERATOR = '0x1111111111111111111111111111111111111111';
+const PAYOUT = '0x2222222222222222222222222222222222222222';
+const NFT_ID = '12345';
+
+const VAULT_POSITION_A = 'pos_vault_owner_a';
+const VAULT_POSITION_B = 'pos_vault_owner_b';
+const NFT_POSITION = 'pos_nft';
+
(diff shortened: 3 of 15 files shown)


0xNedAlbo:
## Architecture review

The work is clean. The identity helper, the fanout DLX, unifying the handlers instead of forking a vault sibling, and making the collaborators injectable are all the right calls — the `EvmConfig.getInstance()` in the constructor was the real defect, not the test.

Two things I would not merge as they stand.

### Blocker 1 — the old queue does not drain, it fills

The code comment says "letting the old one drain". That will not happen. `business-logic.uniswapv3-process-close-order-events` is `durable`, `autoDelete: false`, and **still bound** to `close-order-events` with both routing patterns. A topic exchange delivers to every matching binding, so the v1 queue now receives every close-order event with no consumer attached. It grows without bound until the disk fills.

This needs an explicit deploy step: unbind or delete the v1 queue. My error when approving Q6 — "drain" presumes a consumer, and there is none. Please correct the comment in the code as well, so the next reader does not inherit the assumption.

### Blocker 2 — the poller lost its per-event isolation

`if (!event || !event.nftId) continue;` used to skip an unusable event. `closeOrderRoutingKeyForEvent` now throws when an event carries neither identifier, and in `polling/uniswap-v3-closer.ts` the call sits in the loop with no guard, so one malformed event takes the rest of the batch down with it. The catch-up paths wrap the call per event in `try/catch` — correct there, missing here. Restore skip-and-continue.

### Should be addressed

**The prod deletion is load-bearing, not hygiene.** The upsert keys on `orderIdentityHash`, but the collision risk in R1 sits on `@@unique([positionId, closeOrderHash])`. If legacy lowercase rows exist in prod, the upsert does not find its key, inserts a second row, and then dies on the *other* constraint — dead-lettered, permanently. The upsert protects against the race between the two writers, not against the old data. The deletion must run before or with the deploy, never after, and the PR description should say so in those terms.

**`sharesClosed` is decoded, typed, and then dropped.** It does not reach the log context in `handleExecuted`. AC 5 asks for execution data to be preserved, and for a vault order this is precisely the vault-specific quantity.

### Noted, not for changing

- The DLQ has a seven-day TTL and no dead-letter exchange of its own, so messages disappear silently after a week. What actually satisfies AC 10 is the `error` log line; the queue is retention, not observability. Fine — worth knowing rather than discovering later.
- `findPositionForEvent` queries positions without user scoping, and the ambiguity throw turns a two-user install into a permanent dead-letter. The single-tenant assumption is now load-bearing in a new place.
- Order config stores `nftId` as a string, position config as a number.

Fix the two blockers and the two follow-ups, and this is good to merge. The manual Arbitrum pass should also confirm the dead-letter path end to end, since `onStartup` is the one part of the DLX wiring no test covers — and record whether the reconciliation path produces the row, which is the open Q3.

0xNedAlbo:
Both blockers and both follow-ups addressed — three commits on the same branch, `34de883b`..`c1b5b0d4`.

### Blocker 1 — the v1 queue fills

You're right, and the mechanism is worse than "no consumer": the binding is still live, so from the moment the service moves to `.v2` the old queue starts accumulating *every* close-order event, not just a backlog. `c1b5b0d4` rewrites the code comment to say that, with the `rabbitmqadmin delete queue` line, and the PR description now carries it as a required deploy step rather than cleanup. I also fixed the stale "can be deleted at leisure" phrasing further up the description.

I did not make the rule unbind or delete v1 at startup, and the comment now records why: deleting a queue that may still hold unprocessed v1 messages is destructive, and mid-rolling-deploy is the worst moment for the code to decide that unilaterally. If you'd rather have it automatic, say so and I'll add it — it's a three-line change either way.

### Blocker 2 — per-event isolation

Confirmed and fixed in `34de883b`. Worth noting the blast radius was larger than losing the rest of the batch: the throw escapes `poll()`, so `setCloseOrderLastProcessedBlock` never runs and the cycle's block tracking doesn't advance either. The loop now mirrors the catch-up publishers — skip, log, keep going.

### sharesClosed

`9941038d`. Optional on `OrderExecutedContext`, and omitted entirely for NFT executions rather than written as a placeholder. Two test assertions: present for a vault execution, absent from the context for an NFT one.

### The prod deletion

Reframed as load-bearing, in your terms — the upsert keys on `orderIdentityHash` and cannot see a legacy lowercase row, so it inserts a second row for the slot and then dies on `@@unique([positionId, closeOrderHash])`, dead-lettered and dead-lettered again on every replay. The description now says before-or-with the deploy, never after, and why deleting is safe (the rows are derived; `refresh()` rebuilds them).

### On the three noted items

Taken as read, not changed. Two of them I'd like to name properly rather than leave in a review thread — the DLQ retention/observability distinction is a design fact worth having written down, and the unscoped `findPositionForEvent` plus the ambiguity throw is a single-tenant assumption that is now load-bearing in a place it wasn't before. Want those as issues alongside #79/#80/#81, or are you content with them living here? The `nftId` string-vs-number split I'd fold into #81, since it's the same "two writers, two conventions" shape.

Green after the split: `pnpm typecheck` 14/14, `pnpm build` 11/11, business-logic 36 tests, services 199 tests.

Manual Arbitrum pass now also covers the dead-letter path end to end, since `onStartup` is untested — added as step 4. Still owed, along with the Q3 reconciliation observation. Not merging.

0xNedAlbo:
## Re: fixes

All four addressed as intended. Three notes on your findings and an answer on the issues.

**Blocker 1 — agreed, keep it out of code.** Deleting a queue that may still hold unprocessed messages is a decision that wants a human looking at the queue depth, and mid-deploy is the wrong moment for code to make it. One refinement to the deploy step: make it two phases rather than one delete. Unbind first — that stops the accumulation immediately and is fully reversible — then look at the depth, then delete. If anything is sitting in there, whoever runs the step can decide to shovel it into `.v2` instead of discarding it. A single `delete_queue` line collapses a reversible action and an irreversible one into one command, and the irreversible half is the one that needs judgement.

**Blocker 2 — that is the more serious version of the bug.** A throw escaping `poll()` means `setCloseOrderLastProcessedBlock` never runs and the cycle's block tracking stops advancing. So one malformed event does not lose a batch, it stalls the fallback path indefinitely, and it looks exactly like nothing happening. Same failure family as the original defect: silent, and indistinguishable from a quiet system. Worth remembering as a pattern when reviewing the other pollers later.

**The prod deletion framing** reads correctly now.

**On filing issues — yes, both.** Two different shapes, so keep them separate:

- The DLQ retention-versus-observability gap is a real design question, not a note: nothing surfaces dead letters to the user, and after seven days the evidence is gone. Whether that matters is a product call, so it deserves an issue where it can be weighed rather than a line in a review thread.
- The unscoped `findPositionForEvent` is not a defect to fix; it is an assumption to write down. File it as documenting the single-tenant assumption where it is now load-bearing, so a future reader does not discover it by hitting the ambiguity throw.

Folding the `nftId` string/number split into #81 is right — same "two writers, two conventions" shape as `sharedContractId`.

**Before merging:** please run the manual Arbitrum pass on this branch and post the results here, including step 4 and the Q3 observation. AC 8 and the DLX wiring in `onStartup` are the two parts no test covers, and both deploy steps are load-bearing. I would rather have them confirmed against a chain than merge on green CI alone.

0xNedAlbo:
Two commits since the last round, plus the manual pass — partly done, partly blocked. Being explicit about which is which.

`b9891712` — deploy step split into unbind → check depth → delete, with the shovel option named. You're right that collapsing a reversible and an irreversible action into one command hides the half that needs judgement. Code comment and PR description both updated.

`38ea1087` — new test, see below.

## Step 4 — dead-letter path: PASS ✅

Ran against the dev broker using the **real** rule: real `onStartup` (so the real topology declarations), real `consume`, real `handleMessage` nack path. Collaborators injected as stubs that throw if touched — nothing called them, which also confirms the unroutable path reaches no database.

```
rule started — topology declared by the real onStartup()
v2 queue : args={"x-dead-letter-exchange":"close-order-events-dlx","x-queue-type":"classic"} durable=true
dlq      : args={"x-message-ttl":604800000,"x-queue-type":"classic"} durable=true
v2 bindings : close-order-events -> closer.*.*.*
              close-order-events -> closer.vault.*.*.*
dlq bindings: close-order-events-dlx -> (fanout)

published uninterpretable event, routing key closer.vault.42161.0x13d13B15…5A40.0
RESULT: dlq=1 v2=0
PASS — dead-lettered, not acked and not requeued
```

Test message purged afterwards; DLQ back to 0.

## Steps 1–3 — AC 8 and Q3: blocked, not skipped

Both need an on-chain registration signed by the share holder `0x31bB7b5b7ddf4ad0fc90261Dffe655CDdAeed941`. I have no key for that wallet and won't ask for one, so I can't run them.

Before reporting that, I checked whether existing chain state could substitute — it can't:

- `getOrder(vault, owner, triggerMode)` on the closer for both Arbitrum vault positions × both trigger modes, at block 492,389,530: **all four `status=0` (NONE)**. No live vault order to pick up.
- `eth_getLogs` on `0x13d13B15…5A40` swept back 4M blocks (~11 days of Arbitrum): **zero logs**. No historical registration to replay either.

So the pass genuinely requires a new transaction. Over to you for steps 1–3 — happy to drive the DB and log side while you hold the wallet, if that's easier than doing it solo.

(Aside: the `blockchain` MCP and `cast` both failed here with `Device not configured (os error 6)`, so those reads went through viem with `RPC_URL_ARBITRUM`. Flagging since the repo rule prefers the MCP — looks environmental, not a code problem.)

## What I converted into a test rather than leaving as a claim

`38ea1087` covers the *mechanism* behind AC 8, which had no coverage at all: `apps/midcurve-onchain-data` has no test infrastructure, but the parts that matter live in `@midcurve/services`. It encodes a real vault `OrderRegistered` log from the ABI and asserts the vault topic0 differs from the NFT event of the same name — the precise reason an NFT-only `eth_getLogs` filter excluded every vault event — plus that the decode carries vault, owner and shares, and that the routing key takes the vault shape.

This does not replace the chain pass. It does mean the failure that caused the bug is now pinned by a test instead of by my reading of the ABI.

## One thing the chain reads turned up, relevant to Q3

`shared_contracts` has `UniswapV3VaultPositionCloser` **only for chain 42161**, but there is a vault position on chain 1 (`uniswapv3-vault/1/0x54578F33926F783E9f5E1A994EcCF7a7dcF40B00/…`). For that position `closeOrderService.refresh()` throws `No UniswapV3VaultPositionCloser contract found for chain 1`, and `UniswapV3VaultPositionService.refresh()` swallows it as a `warn`.

That is a concrete instance of exactly what Q3 was asking about, found without a transaction: on mainnet the reconciliation path is silently a no-op. It doesn't change this PR — the event path now handles chain 1 fine once a closer is registered there — but it means "the wizard refresh will create the row" is only true on Arbitrum today. Worth knowing before the manual pass, and worth deciding whether the swallowed catch should distinguish "no contract deployed on this chain" from a real failure. Say the word and I'll file it.

## Issues filed

- #82 — dead-lettered events are not surfaced anywhere and expire after 7 days, framed as the product call you described
- #83 — documenting the single-tenant assumption where it is now load-bearing, not a fix
- the `nftId` string/number split folded into #81

Green: typecheck 14/14, build 11/11, business-logic 36 tests, services 204 tests (5 new). Not merging.

0xNedAlbo:
## Re: verification results

Step 4 is convincing — real `onStartup`, real consume, real nack, and the observation that no stub was called is the part that matters: it shows the unroutable path touches no database before dead-lettering.

**The refusal on steps 1–3 was the right call**, and worth stating plainly so it is not read as a gap in diligence. Asking for a signing key to complete a test is a line that should not be crossed, and the alternative — presenting existing chain state as evidence — would have been worse. Checking whether the steps could be satisfied without a transaction (`getOrder` returning `status=0` for both positions × both modes, 4M blocks with no logs from the closer) and then reporting "blocked" rather than "passed" is exactly the behaviour this verification needs. Jan holds the wallet and will run the registration; drive the DB and log side from there.

**The chain-1 finding is more important than this PR.** It is the same failure mode we just removed from the rule's `default` branch: an outage that looks like a quiet system. Two separate pieces of work, please file both:

1. **The swallowed catch is the bug.** `UniswapV3VaultPositionService.refresh()` downgrades *every* close-order refresh failure to a `warn`, regardless of cause or chain. A missing contract registration, an RPC outage and a genuine defect are currently indistinguishable, and all three look like success. That is a pattern, not one bad line — file it as such, scoped to the swallow, not to the chain.

2. **The missing registration is a deployment task, and the answer is: deploy everywhere.** The vault closer should be installed and registered on all three production chains — Ethereum (1), Arbitrum One (42161) and Base (8453). It exists only on 42161 today, so Ethereum and Base are outstanding. Not a documented limitation, and not part of this PR.

The step-8 test is well placed. Pinning that the vault topic0 differs from the NFT one is precisely why the old filter excluded every vault event — a claim has become a guarantee.

**Merge recommendation unchanged.** The unit tests cover the seam that was broken, but steps 1–3 exercise the chain across receipt publisher, rule, monitor and executor — and that is the class of defect this issue was about from the start: every layer sound, the wiring cut. One transaction on Arbitrum settles it.

0xNedAlbo:
One of the two filed, one already done — checked before filing rather than after.

## 1. The swallowed catch → #86

Filed, scoped to the swallow as you asked. It is a pattern and it is wider than vault: **three instances, both protocols**.

| Location | Call |
|---|---|
| `uniswapv3-vault-position-service.ts:583` | `refresh()` reconciliation |
| `uniswapv3-position-service.ts:1346` | `refresh()` reconciliation |
| `uniswapv3-position-service.ts:899` | `discover()` during position import |

All three downgrade every cause to a `warn` and return as if nothing happened, so the caller cannot distinguish "reconciled, no orders" from "could not reconcile". The issue quotes the chain-1 case as evidence of what the swallow hides rather than as the thing to fix — the next instance will look identical and be equally invisible.

## 2. The missing registration — already deployed, nothing to file

Re-checked the registry before writing the issue, and it has changed since my earlier read this morning. The vault closer is now registered on all three production chains:

| Chain | Address | Registered |
|---|---|---|
| Ethereum (1) | `0x1b373BA92c24e1aCA38bC0d78CB64209F81DDe2E` | today 12:57 |
| Arbitrum (42161) | `0x13d13B15BbE9b06C0279a7aB5f0a898EA3f25A40` | 2026-04-10 |
| Base (8453) | `0x66deed7C4669680BEF6484E269689490aA385d9a` | today 13:05 |

Verified rather than taken from the registry: all three have bytecode deployed (191 bytes, consistent with the proxy pattern), and both new ones answer `getOrder(address,address,uint8)` with a decodable `VaultCloseOrder` tuple — so they are vault closers responding to the ABI we actually call, not just addresses with code at them.

The two new rows landed at 12:57 and 13:05, i.e. after your review at 12:35. So the deployment task was done while this was in flight, and filing "deploy everywhere" would have been filing completed work. Say the word if you want a tracking issue anyway for the record.

**One remaining delta, your call whether it matters:** the NFT closer is registered on 5 chains (1, 8453, 42161, plus Sepolia 11155111 and local 31337); the vault closer on 3. Production is complete either way — Sepolia and local only matter if vault close orders are meant to be exercisable in those environments. Happy to file that narrowly if it is wanted.

## Status

Nothing else changed on the branch — no code in this round, so the head is still `38ea1087`. Steps 1–3 remain owed and are Jan's transaction to make; I will take the DB and log side once they run it. Not merging.

Issue #88, opened 2026-08-08
Close-order catch-up: oversized scan batching, cursor advancing past unscanned blocks, and backfill blocking live polling


0xNedAlbo:
**Body rewritten 2026-08-13.** The three defects were filed after a measured run on 2026-08-08. Two of them have since been fixed or removed by other work, one supporting claim was wrong, and the decision on batch sizing has changed the shape of the rest — the fix is now mostly a deletion. The original measurements and defect analysis are kept below where they still hold.

Found while verifying #77 on Arbitrum.

## The measurement that started this

Dev stack, 2026-08-08, one restart of `midcurve-onchain-data`:

| chain | blocks scanned | `eth_getLogs` calls | duration | events found |
|---|---|---|---|---|
| 1 | 673,088 | ~68 | 99 s | 0 |
| 8453 | 4,053,747 | ~406 | **7 m 43 s** | 0 |
| 42161 | — | — | still running at 8 min | — |

No poller existed for any chain during that window. That is what blocked AC 8 of #77 from being verified: the order was registered on-chain, and nothing was listening.

## Status of the original three defects

**Defect 2c — the heartbeat advancing the cursor with no scanning — is fixed.** `updateBlockTrackingHeartbeat()` now persists `poller.getLastScannedBlock()` through `updateCloseOrderBlockIfHigher`, and the method carries a comment stating that the chain head is deliberately not read here because a restart would otherwise resume past unscanned blocks. Nothing to do.

**One site of Defect 2b is gone** — the non-finalized publish path was deleted with #93.

**Defect 1, the remaining sites of 2a and 2b, 2d and Defect 3 all still hold** against `main`, with line numbers drifted.

## Decision: one sweep per chain, no batching

`CATCHUP_BATCH_SIZE_BLOCKS` and `CLOSER_POLL_BATCH_SIZE` both default to `10000`. That window is removed, not enlarged: scan `fromBlock → toBlock` in a single `eth_getLogs` call.

This rests on assumptions we are choosing to bake in rather than detect:

- **Alchemy as the RPC provider**, on a **Pay As You Go** plan or better.
- **Ethereum, Arbitrum and Base only** — the three chains this deployment supports, all in Alchemy's unlimited-range set.
- Responses are capped at 150MB. A closer-address-and-topic filtered query returning zero-to-few logs is nowhere near it, even over full chain history.

[...]

Issue #79, opened 2026-08-08
Automation log duplicates entries when config-update events are replayed


0xNedAlbo:
**Body extended 2026-08-13.** The problem analysis is unchanged and verified against `main`. Added: the decision on what the key covers, the migration's treatment of existing rows, and two traps found while reading the code that determine whether the obvious implementation works at all.

Split out of #77 (see the [decisions comment](https://github.com/0xNedAlbo/midcurve-finance/issues/77#issuecomment-5225632075), R4). Affects `uniswapv3` and `uniswapv3-vault` equally.

## Problem

`AutomationLogService` only ever `create`s — there is no dedupe key. Verified: `log()` in `packages/midcurve-services/src/services/automation/automation-log-service.ts` calls `db.automationLog.create()` and nothing else, and all thirteen convenience methods route through it.

The lifecycle handlers are replay-safe by accident in two of three cases:

- **Registration replay** — an already-`monitoring` order short-circuits before any log is written. Safe.
- **Cancel / execute replay** — the order is already deleted, so `resolveOrder` returns `null` and nothing is logged. Safe.
- **Config-update replay** — `mergeState` is idempotent, but `logOrderModified` appends a fresh row every time. **Not safe.**

## How replays actually occur

The original text said replays happen because the non-finalized catch-up republishes by design. **That is wrong**, and as of #93 that code no longer exists at all — the non-finalized catch-up has been deleted and the finalized-only scan plus the poller is what we ship.

The duplication is nonetheless reachable today, by a narrower route: Defect 2d in #88 — `fromBlock = cachedBlock` with an inclusive `toBlock` — rescans the cursor boundary block on every startup. A config update landing in exactly that block replays and appends a second `ORDER_MODIFIED` row. Rare rather than routine.

It becomes routine once #88's cursor fix lands, since holding the cursor back over a partially-failed range republishes everything in that range on retry. That is correct at-least-once behaviour, and this issue is what makes it safe to ship.

## Effect

The automation log an operator reads to reconstruct what happened to an order accumulates duplicate `ORDER_MODIFIED` entries after every pass over a block range that contained a config update.

### The key covers every log write that originates from an on-chain event, not only `ORDER_MODIFIED`

The issue's own analysis is that the other two paths are safe **by accident** — one short-circuits on state, one finds a deleted row. Both arguments are about the current shape of unrelated code, and #88 is about to change how often events are replayed.

[...]

PR #90, opened 2026-08-08
fix(onchain-data): persist only scanned blocks in the close order heartbeat

### LINKED ISSUE #89: Block-tracking heartbeat advances the persisted cursor without scanning, skipping up to an hour of blocks on restart
Split out of #88, where this is Defect 2c. Separated because it is the only one of that issue's three defects that loses production data unconditionally, while the other two are cost and startup-ordering problems. Bundled, it cannot be prioritised or shipped on its own.

## The defect

`updateBlockTrackingHeartbeat()` writes the persisted cursor on a timer, without scanning anything:

`setCloseOrderLastProcessedBlock(chainId, currentBlock)` — [`workers/uniswapv3/uniswapv3-close-order-poller.ts:325`](apps/midcurve-onchain-data/src/workers/uniswapv3/uniswapv3-close-order-poller.ts#L325), fired every `heartbeatIntervalMs` (`CATCHUP_HEARTBEAT_INTERVAL_MS`, default 60000).

`currentBlock` is the chain head. No `eth_getLogs` call is made for the range between the last scanned block and that head. The cursor simply moves.

## Why it loses events

The running poller keeps its own in-memory `lastProcessedBlock`, so live polling is unaffected while the process stays up. The damage is at restart: the next startup reads the *persisted* cursor, which the heartbeat has advanced to wherever the chain head was 60 seconds ago, and begins scanning from there. Everything between the last actual scan and that point is never looked at.

With the default poll interval of one hour (`CLOSER_POLL_INTERVAL_MS`, 3600000), a restart can therefore skip up to an hour of blocks on every configured chain. Silently — no error, no warn, no gap in any log.

This is worse than the other two cursor problems in #88 in one specific way: 2a and 2b need something to fail first. This one needs nothing. It happens on every heartbeat of every healthy run.

## Scope

Two candidate shapes, both acceptable:

- Delete the cursor write from the heartbeat entirely. The heartbeat's purpose is liveness signalling; cursor persistence belongs to whatever actually scanned a range.
- Or have it advance only to the highest block that was in fact scanned and published — which makes it a persistence of the in-memory value rather than a jump to the head.

The second is closer to the apparent intent and keeps the restart cursor warm. Either removes the loss.

## Out of scope

The other two cursor defects from #88 (2a: a failed batch skipped with the cursor advancing over the whole range; 2b: a failed publish skipped with the cursor still advancing) stay in #88, along with the batching and startup-ordering work. Fixing them properly means holding the cursor back over partially-failed ranges, which raises replay volume and therefore depends on #79. **This issue has no such dependency** — removing an unconditional cursor jump does not increase replay, it only stops skipping. It can land on its own, ahead of both.

## Footgun: `CATCHUP_ENABLED` switches off two things

Worth stating in the body rather than leaving in a comment thread, because the situation that makes someone reach for it is the one where it does the most damage.

`startBlockTracking()` returns early on `config.enabled` ([`workers/uniswapv3/uniswapv3-close-order-poller.ts:287`](apps/midcurve-onchain-data/src/workers/uniswapv3/uniswapv3-close-order-poller.ts#L287)), so `CATCHUP_ENABLED=false` disables the startup catch-up **and** the block-tracking heartbeat together. Nothing in the variable's name suggests the second.

That matters here specifically: disabling catch-up is the obvious reaction to the slow startup described in #88, and once this issue is fixed the heartbeat is also the only thing that repairs a cursor write that failed inside `poll()` — those failures are swallowed and warned, never retried. Turning catch-up off to make startup faster therefore silently removes the repair path for the cursor as well.

The fix for a slow startup is `CATCHUP_BATCH_SIZE_BLOCKS`, not `CATCHUP_ENABLED`.

## Verification

A restart after an idle period longer than one heartbeat should resume scanning from the last block actually scanned, not from the chain head at the time of the last heartbeat. Worth asserting on the persisted value directly rather than inferring it from event counts, since the whole failure mode is that no events appear either way.

`CATCHUP_HEARTBEAT_INTERVAL_MS` and `CATCHUP_ENABLED` are both read by `getCatchUpConfig()` and neither is documented in `.env.example`, unlike the batch-window variables. Worth adding while in here — an operator tuning the scan window currently has no way to learn that a second timer is also moving the cursor.

0xNedAlbo:
Closes #89

The block-tracking heartbeat wrote the chain head to the persisted cursor every 60 s without scanning anything. Live polling was unaffected — the poller keeps its own in-memory position — but a restart read the *persisted* value and resumed from wherever the head had been at the last heartbeat, skipping up to `CLOSER_POLL_INTERVAL_MS` (default 1 h) of blocks on every configured chain, with no error and no gap in any log.

Implemented per the [agreed decisions](https://github.com/0xNedAlbo/midcurve-finance/issues/89#issuecomment-5226757049): option 2, `hasCompletedScan` gate included, vitest added, `.env.example` gets its own subsection.

## What changed

**The heartbeat now re-persists what was scanned, not where the chain is.** It reads each poller's watermark and writes it through `updateCloseOrderBlockIfHigher()` — which already existed for exactly this purpose and had zero callers. It no longer touches the chain at all, dropping one `eth_getBlockNumber` per chain per minute that existed only to produce the value being mis-persisted.

**The watermark is gated behind a completed poll cycle.** A poller with no cached cursor seeds its position from the chain head, and nothing has been scanned at that moment. Publishing that value would reintroduce a bounded version of the same defect — and, more to the point, would make this fix's correctness depend on catch-up having succeeded.

The gate does not remove that dependency entirely, and the code comment now says so: on a **cold start** (no cached cursor **and** catch-up disabled) `hasCompletedScan` stays false for one poll interval, nothing is persisted, and a restart inside that window re-seeds from the head. The same skip, bounded to one interval rather than unbounded, and unreachable in normal operation since catch-up writes the cursor before any poller starts. What the gate does buy is that every *subsequent* restart is independent of catch-up.

**Kept deliberately:** the cursor still advances over ranges where individual publishes failed. That is #88's 2b and stays there.

## Test

`apps/midcurve-onchain-data` had no test infrastructure; this adds vitest mirroring `midcurve-business-logic` (one new devDep, a 6-line config, two scripts — no `turbo.json` change).

The tests assert on the **persisted cursor value**, not on call arguments, because the failure mode is that an unscanned range and an empty one are indistinguishable from the outside. `@midcurve/database` is mocked with an in-memory cache table so the real `CacheService` and the real cursor helpers run against it.

Verified as a regression guard: all seven fail against the pre-fix heartbeat, with `expected 250000000n to be 249000000n` — the head persisted in place of the scanned block.

```
✓ persists the scanned watermark, never the chain head
✓ does not read the chain head at all
✓ advances the cursor when the poller has scanned further
✓ never moves the cursor backwards
✓ writes nothing for a poller that has not completed a scan
✓ leaves the cursor untouched when there is no cursor and no scan
✓ reports no scanned block while seeded from the chain head
```

`typecheck` and `build` pass. `pnpm lint` fails on this app both before and after this branch — it has no `eslint.config.js` at all, which is pre-existing and out of scope here.

## CI

`pr-tests.yml` filtered to three packages, so the suite above would have run on my machine and never again. This PR adds `services` (204 tests), `business-logic` (36) and `onchain-data` (7) to the workflow — 247 tests that were previously invisible to pull requests.

The services scripts hardcoded `./node_modules/.bin/vitest`, which does not exist under the repo's `node-linker=hoisted`, and loaded a gitignored `.env.test`. Both are dropped; the suite is green in a bare environment with no RPC URLs, no `DATABASE_URL` and no `.env`. That is a change to another package's manifest inside this PR — flagging it rather than burying it.

`@midcurve/signer` is deliberately **not** added: no `vitest.config.ts`, no test files of its own, so `vitest run` falls back to the default glob and collects UI Playwright specs out of `.next/standalone` — 112 failing files. Vestigial script, wants its own issue.

## Railway config

Raising `CATCHUP_BATCH_SIZE_BLOCKS` is still worth doing, but it is **not** a merge gate and my earlier framing of it as one was wrong. The reasoning that corrected it: on a healthy restart the widened range is bounded by the poll interval, and on an unhealthy one by the downtime — the same bound as before, since the old heartbeat also left the cursor wherever the head was when the process stopped. The 7m43s in #88 came from a cursor millions of blocks stale, which is a different situation. Do the config change because startup ordering is broken (#88 Defect 3), and merge this when CI is green.

No new replay: the blocks added to the scan range are ones never scanned, so their events publish for the first time rather than again. The #79 dependency argument in #88 does not apply here.

## Manual verification

```bash
psql $DATABASE_URL -c "select key, value, \"updatedAt\" from cache \
  where key like 'onchain-data:close-order-subscriber:last-block:%' order by key;"
```

Sampled twice ~90 s apart with the poll interval at its 1 h default: `blockNumber` holds steady while `updatedAt` advances. On `main`, `blockNumber` moves.


0xNedAlbo:
## Review

The substance is right. Option 2 with the `hasCompletedScan` gate is the correct shape: it separates "where we started looking" from "what we have looked at", which is exactly the distinction the defect collapsed. Dropping the per-chain `eth_getBlockNumber` is a real bonus — the call existed only to produce the wrong value. Reusing `updateCloseOrderBlockIfHigher` rather than adding a monotonic guard is the better find, and keeping #88's 2b out of scope is correctly drawn.

One thing needs to happen in this PR.

### The new tests do not run in CI

`.github/workflows/pr-tests.yml` runs `build`, `typecheck`, and then `test:run` for exactly three filtered packages: `@midcurve/shared`, `@midcurve/api-shared`, `@midcurve/mcp-server`. Nothing else. Adding a `test` script to `apps/midcurve-onchain-data` does not put it in front of a pull request — the seven tests here will pass on your machine and never run again.

The PR description lists "no `turbo.json` change" as a plus. It is in fact the mechanism by which the suite stays invisible. Please add the workflow step in this PR; a regression guard that does not run on PRs is a comment with a test framework attached.

While you are in that file: the same applies to work already merged. The 23 business-logic tests from #78 and the services suite are both outside the filter list, so "green" on those PRs meant green locally. Add steps for them too if they run clean in CI, or tell me which do not and I will scope it separately — but the gap should not survive this PR unexamined.

### The Railway precondition should not gate the merge

Raising `CATCHUP_BATCH_SIZE_BLOCKS` is worth doing, but the reason is #88 Defect 3, not this change. The scan range this fix widens is bounded by the poll interval on a healthy restart, and by the downtime on an unhealthy one — the same bound as before the fix, since the old heartbeat also left the cursor at wherever the head was when the process stopped. The 7m43s in #88 came from a cursor millions of blocks stale, which is a different situation. Do the config change because startup ordering is broken, and merge this when CI is green.

### One line worth adding to the comment

The gate has an edge worth naming where the next reader will be: with no cached cursor **and** catch-up disabled, `hasCompletedScan` stays false for a full poll interval, so nothing is persisted and a restart inside that window re-seeds from the head — the same skip, bounded to one interval. Not a regression, and in normal operation catch-up writes the cursor before any poller starts. But it means the cold-start cursor still depends on catch-up having run, and a reader who takes the independence claim in the PR description literally will not expect that.

Everything else — the test design asserting on the persisted value rather than call arguments, reaching into private state for a timer-driven method, the `.env.example` subsection including the `CATCHUP_ENABLED` footgun — is well judged. The `eslint.config.js` absence is pre-existing and correctly left alone; worth its own issue rather than a fix here.

0xNedAlbo:
All three addressed — `71f8c2b8` (CI) and `d9c43b0e` (comment), plus a rewritten PR description.

## CI: you were right, and it was worse than one package

I examined the gap rather than just adding my own step. Four packages had test scripts outside the filter; three are now in the workflow:

| package | tests | status |
|---|---|---|
| `@midcurve/services` | 204 | added |
| `@midcurve/business-logic` | 36 | added |
| `@midcurve/onchain-data` | 7 | added |
| `@midcurve/signer` | — | **not** added, see below |

247 tests that never ran against a PR. All six suites now green locally in a bare environment — no RPC URLs, no `DATABASE_URL`, no `.env` — so nothing here depends on a secret CI does not have. `business-logic` is 36 rather than the 23 you remembered; there are two test files now.

**One thing needs your eye, because it touches another package.** `@midcurve/services` did not run at all, in CI *or* locally, for two reasons: its scripts hardcode `./node_modules/.bin/vitest`, which does not exist under the repo's `node-linker=hoisted`, and they load a gitignored `.env.test`. I changed `test` and `test:run` to plain `vitest` — the suite passes 204/204 without the env file, so it was loading secrets it never used. I left `test:coverage` and the `:integration` variants on the old form, unverified and untouched. Revert that hunk if you would rather scope it separately; the workflow step goes with it.

`@midcurve/signer` I left alone deliberately. It has no `vitest.config.ts` and no test files of its own, so `vitest run` falls back to the default glob, crawls `.next/standalone` build output and collects the UI's Playwright specs — 112 failing files. The script is vestigial rather than broken-but-close. Worth its own issue alongside the `eslint.config.js` one; say the word and I will file both.

## Railway: agreed, my framing was wrong

Your bound is the correct one. The old heartbeat also left the cursor at the head when the process stopped, so downtime dominates in both cases and this fix adds at most one poll interval on top — not the millions-of-blocks-stale situation that produced the 7m43s. I had reasoned from "the range gets wider" to "this gates the merge" without checking what the delta actually was. The PR description now says raise it because of #88 Defect 3, and merge on green CI.

## The gate's cold-start edge

Added where you said it belongs, on the `hasCompletedScan` field, and the independence claim in the PR description is qualified to match. The distinction the comment now draws: the cold-start cursor still depends on catch-up having run; every *subsequent* restart does not. That is a narrower claim than the one I made, and the accurate one.

Nothing else changed — the fix, tests and `.env.example` are as reviewed. Not merging.

0xNedAlbo:
## Re: CI gap and the services suite

**Keep the `@midcurve/services` hunk in this PR.** The workflow step is worthless without it, and splitting a two-line script fix from the step that makes it matter would leave main in a state where the suite is wired into CI and cannot execute. That is worse than either end state.

**One check before we treat 204/204 as a clean bill.** "Passes without the env file, so it was loading secrets it never used" does not follow from the pass count alone. A suite that gates on missing configuration usually does it by skipping, not by failing — `describe.skipIf`, an early `return`, a guard in `beforeAll`. Since the suite could not run at all before, there is no baseline count to compare against, so 204 passing is consistent both with "nothing needed the env" and with "the tests that needed it quietly skipped themselves".

Please report the skipped and todo counts alongside the passing count, and grep the suite for conditional skips keyed on environment variables. If some tests do skip without `.env.test`, that is fine and probably correct — but then CI is running a subset, and we should know which subset rather than believe we have 204.

**`@midcurve/signer` — right call to exclude, wrong end state to leave it in.** A script whose only effect is to collect 112 Playwright specs out of `.next/standalone` and fail is a trap for whoever next runs tests at the workspace root. Deleting the script is smaller than the issue describing it, and removes the trap rather than documenting it. Do that here.

Worth noting separately what it exposes: a build artifact directory is being globbed as source. That is worth a look on its own terms, independent of the script.

**The filter list is the underlying problem.** A hand-maintained enumeration of packages in the workflow drifted for four packages over months, silently, and would drift again the next time someone adds a suite — this PR fixes the instances, not the mechanism. The durable shape is running tests across the workspace so a new package with a `test:run` script is picked up by existing, with explicit opt-outs where a package genuinely has no suite. The signer case is exactly why that cannot be done naively: packages without a vitest config fall back to the default glob and collect whatever is lying around. Not for this PR — but file it, because "we enumerated correctly this time" is not a fix.

**On Railway and the cold-start edge:** both settled, nothing further from me. The reasoning-from-shape-to-conclusion you describe is the same failure the heartbeat had — a plausible chain that nobody checked against the actual delta.

**Issues to file:** the workspace-wide test discovery above, and the missing `eslint.config.js`. Not the signer script — delete it here instead.

0xNedAlbo:
All four done. `d8508a73` deletes the signer scripts; issues #91 and #92 filed.

## The skip check — you were right to demand it, and the answer is clean

My "so it was loading secrets it never used" was inferred from a pass count, which is exactly the reasoning shape you flagged. Here is the evidence instead.

**A baseline does exist** — I could run the original invocation with a working vitest path, `.env.test` and all:

| invocation | files | tests |
|---|---|---|
| `vitest run` (bare env, no `.env.test`, `ETHERSCAN_API_KEY` unset) | 13 passed | **204 passed** |
| `node --env-file=.env.test vitest run` (secrets loaded) | 13 passed | **204 passed** |

Identical, so nothing appears or disappears with the env file present.

**Zero skipped, zero todo.** Vitest prints those in the summary when non-zero (`Tests 200 passed | 4 skipped`); the line reads `Tests 204 passed (204)` in both runs. Confirmed with `--reporter=basic`.

**No mechanism for a conditional skip exists in the suite.** `grep -rnE "skipIf|runIf|\.skip\b|\.todo\b"` across all 13 test files: no hits. `grep -rn "process\.env"` across the same: no hits. So there is no `describe.skipIf`, no guarded `beforeAll`, no early return keyed on configuration — the tests never look at the environment at all.

**And it now has a real CI run behind it.** The workflow went green on GitHub's runners, where no `.env.test` exists and none of those secrets are set: `Test @midcurve/services` ✅ along with all five others.

There are also no integration test files — the `coingecko-client.integration.test.ts` in the services CLAUDE.md does not exist on disk. That doc is stale, which is worth knowing but not something I touched here.

## Signer

Scripts deleted, and the CLAUDE.md dev commands pointed at `pnpm test`, so that line now points at `test:api` — the manual API smoke script, which is unrelated to vitest and stays.

With those gone, the six packages exposing `test:run` are exactly the six the workflow runs, verified by enumerating the workspace. I left the unused `vitest` devDependency in place: removing it churns the lockfile for no behavioural gain, and it is what a real suite there would need. Say so if you would rather it went.

## Issues filed

- **#91** — workspace-wide test discovery. Argues `pnpm -r test:run` over the hand-maintained list, and records why it cannot be done naively: the signer glob hazard is unchanged for any *future* package that gains a `test:run` without a config bounding `include`. Also carries your point about `.next/standalone` containing copies of other packages' source, as a hazard worth its own look.
- **#92** — the ESLint gap, which is **five packages, not one**. `@midcurve/api`, `@midcurve/automation`, `@midcurve/business-logic`, `@midcurve/onchain-data` and `@midcurve/signer` all declare `"lint": "eslint ."` with no config anywhere, and there is no root config to inherit. `@midcurve/ui` is the only package in the repo that is actually linted, and has been since the ESLint 9 flat-config default landed.

## On the reasoning failure

Taken. Both times it was the same move — a plausible chain accepted without checking the quantity it depended on. Worth noting the test I added has the property that would have caught the heartbeat version of it, and the numbers above are the version of that for the claim about `.env.test`.

CI is running on the new commit. Not merging.

0xNedAlbo:
The skip check is convincing — a real baseline, identical counts both ways, no skip mechanism present in any of the 13 files, and green on runners that hold none of those secrets. That is the evidence the claim needed.

**Leave the `vitest` devDep.** Not worth a lockfile change on its own; it can go with whatever next touches that package.

**The stale `coingecko-client.integration.test.ts` reference** is worth correcting while the services CLAUDE.md is fresh in mind — a documented test that does not exist is the same shape as a workflow that lists packages it does not run.

**One question on #92 before it gets picked up.** Five packages declare `"lint": "eslint ."` with no config and no root config to inherit. Under ESLint 9 that is an error, not a no-op — so either `pnpm lint` at the workspace root has been failing for months and the failure is routinely ignored, or the script is never invoked at all. Those are different problems: the first is a red check people have learned to look past, which is worse than no check; the second is dead configuration. Worth establishing which before deciding what the fix looks like, and worth recording in #92 either way.

That makes three instances in this PR of the same shape — a guard that appears to be in place and is not running: the workflow's package filter, these lint scripts, and the heartbeat that looked like it was tracking scanned blocks. Different mechanisms, same failure to verify that the verification runs.

0xNedAlbo:
CI green on `d8508a73`, the signer-cleanup commit — [run 31268204895](https://github.com/0xNedAlbo/midcurve-finance/actions/runs/31268204895), 3m52s.

All six test steps pass on GitHub's runners, including the three added here:

```
✅ Build          ✅ Test @midcurve/shared        ✅ Test @midcurve/services
✅ Typecheck      ✅ Test @midcurve/api-shared    ✅ Test @midcurve/business-logic
                  ✅ Test @midcurve/mcp-server    ✅ Test @midcurve/onchain-data
```

That is also the independent confirmation on the `.env.test` question: `@midcurve/services` passes in an environment where none of those secrets exist and no `.env.test` was ever checked in.

Ready for your call. Not merging.

Issue #98, opened 2026-08-09
Roll back the accounting layer


0xNedAlbo:
Phase 2b of the platform reduction, and its centre. Lands **after** #97. The inventory this is based on is the [main comment on #96](https://github.com/0xNedAlbo/midcurve-finance/issues/96#issuecomment-5230757456) with its [addendum](https://github.com/0xNedAlbo/midcurve-finance/issues/96#issuecomment-5230787978) — read both before planning. Section references below point there.

## Decision

The accounting layer is removed: journal entries, token lots, the chart of accounts, and the double-entry posting rules. The platform keeps position management, not bookkeeping.

The `position-ledger` **stays**. §8 established the cut is clean in the direction that matters — no accounting code creates, updates or deletes ledger rows; no ledger field or event type exists solely to be posted; the accounting-only vocabulary lives in a separate enum inside the accounting schema. PnL, cost basis, APR and the ledger UI are all ledger-owned and survive intact.

## Two corrections to the original scope

**"Remove the wallet perimeter" was wrong and must not be followed literally.** Per §9-A, `UserWallet` and `UserWalletService` are load-bearing for the surviving pipeline — four call sites in `UniswapV3PositionService`, the SIWE login path, the wallets API, and a full UI screen. They stay untouched. Removable under that heading is only the dead half, none of which has a caller: `KnownProtocolAddress`, `KnownProtocolAddressService`, `EvmWalletPerimeterService`, and the `TransferClassification` types in `@midcurve/shared`.

**Wallet changes must keep reaching the ledger.** §9-B is the one place where deleting accounting would change the surviving core: `UniswapV3ReevaluateOnWalletChangeRule` is the only consumer of `wallet.added` / `wallet.removed`, and it calls `recalculateAggregates()`, which recomputes `isIgnored` and every running total across the user's positions. That call is ledger maintenance that happened to live in an accounting rule.

**Decision: keep a slimmed rule.** It binds the wallet events, calls `recalculateAggregates`, and posts nothing. Same observable behaviour, accounting removed from around it. A reduction may remove capability the product no longer wants; it may not silently change what the core does.

## What must be gone

Per §1–§7. Summarised rather than restated — the inventory is the working list.

- **Prisma**: the whole `accounting` schema (6 models + 1 enum), `schemas` back to `["public"]`, and the seven back-relation fields on `User`, `PositionLedgerEvent` and `Token`. No FK points from a surviving model into accounting, so removal is additive in that direction.
[...]

PR #99, opened 2026-08-09
refactor: remove the uniswapv3-staking position type

### LINKED ISSUE #97: Remove the uniswapv3-staking position type
Phase 2a of the platform reduction. The full inventory this is based on is the [addendum on #96](https://github.com/0xNedAlbo/midcurve-finance/issues/96#issuecomment-5230787978), Part A — read it before planning. This issue states what must be true afterwards, not how to get there.

## Decision

The `uniswapv3-staking` position type is removed from the tree. Surviving protocols are `uniswapv3` (NFT) and `uniswapv3-vault`, nothing else. #63 parked it as a feature; this removes the source.

The evidence that made this cheap: staking has **zero UI and zero MCP surface**. `position-list.tsx` does not list it in `VALID_PROTOCOL_VALUES` and drops staking positions through `default: return null`, while `POST /api/v1/positions/discover` still scans for and imports them. The system can create positions today that no screen will ever render.

## Sequencing

This lands **before** the accounting rollback. It removes ~1,018 LOC of the accounting problem as a side effect and ~6,195 LOC unrelated to it, leaving a smaller surface for the harder pass. Do not combine the two.

## What must be gone

Per the #96 addendum A.1 — 27 files, ~7,213 production TypeScript LOC, plus tests, Solidity and docs:

- The three services (position, ledger, APR) and their tests
- The staking types in `@midcurve/shared` and `@midcurve/api-shared`
- All nine API routes, including the staking `/accounting` route
- `UniswapV3StakingPostJournalEntriesRule` and its test — the only tested unit in the accounting layer, which is why the business-logic suite drops here rather than during the rollback
- The Solidity contracts, their tests and mocks, and the deploy script
- `docs/positions/uniswapv3-staking-vault/`
- `createStakingShareTokenHash`, `StakingShareTokenHashData`, and the `staking-share` parse case in `token-hash.ts`
- The staking branch in `POST /api/v1/positions/discover`

Wiring points that lose a branch rather than disappearing: the `PositionProtocol` and `LedgerEventProtocol` unions, `PositionFactory` and `PositionLedgerEventFactory` (three sites each), the three service barrels, `serializers.ts` and `services.ts` in the API, `SharedContractType.UNISWAP_V3_STAKING_VAULT_FACTORY`, and `seed-contracts.ts`.

## Database rows outlive the code — handle this first

Deleting the factory branches makes any surviving row with a staking discriminator **uninstantiable**, on tables that are read on ordinary page loads. `PositionFactory.fromDB` will throw where it previously dispatched. Three row sets need checking against production before anything merges:

| Table | Condition | Consequence if left |
|---|---|---|
| `positions` | `protocol = 'uniswapv3-staking'` | position list throws on every load |
| `position_ledger_events` | belonging to those positions | cascade dependency of the above |
| `shared_contracts` | the staking vault factory entry | enum value no longer exists |

Report what production actually holds before deciding the cleanup. If the answer is "none", say so explicitly — that is a result, not an absence of work.

The on-chain deployment is unaffected and stays where it is; `deployments/arbitrum.json` is a historical record and its handling is a judgement call to state rather than make silently.

## Test expectations

From the #95 baseline, per file rather than in total:

- `@midcurve/services`: **204 → 153**. Removed: `uniswapv3-staking-ledger-service` (30), `uniswapv3-staking-position-service` (14), `uniswapv3-staking-apr-service` (6), and one staking smoke case in `position-list-service.test.ts:194`.
- `@midcurve/business-logic`: **36 → 24**. The 12 staking journal-rule cases go; the 24 close-order cases must be untouched.
- Every other suite: **unchanged**. `shared` 127, `mcp-server` 13, `api-shared` 8, `onchain-data` 7.

A drop anywhere outside those two lines is collateral damage and a reason to stop rather than adjust the expectation.

## Acceptance criteria

1. No file in the tree contains a `uniswapv3-staking` protocol branch, and the protocol unions have two members.
2. `POST /api/v1/positions/discover` no longer scans for staking vaults; NFT and vault discovery behave exactly as before.
3. The position list, position detail, close orders, ledger and APR paths behave identically for both surviving protocols — verified by the suites above holding at their stated counts.
4. No database row remains whose discriminator the code can no longer instantiate, on any of the three tables above.
5. The chart of accounts, journal, token lots and both surviving journal rules are **untouched** — they go in the next pass.

## Out of scope

The accounting layer beyond the staking journal rule and the staking `/accounting` route. The `basic-currency` machinery, which is dead on its own account (#93). Anything in #93's sweep. `User.reportingCurrency` and `costBasisMethod`, which belong to the accounting rollback.

0xNedAlbo:
Phase 2a of the platform reduction. Removes the `uniswapv3-staking` position type from the tree; surviving protocols are `uniswapv3` and `uniswapv3-vault`.

Closes #97

Implements the plan as answered in the issue thread — 48 files deleted, 23 edited, **18,577 lines removed** against 21 added.

## What went

| Area | Deleted | Edited |
|---|---|---|
| `@midcurve/shared` | 9 type files | protocol unions, both factories, 3 barrels, `token-hash.ts`, `SharedContractName` |
| `@midcurve/api-shared` | 4 response-type files | `positions/index.ts` |
| `@midcurve/services` | 3 services + 3 test files | 3 barrels, `position-list-service.test.ts` |
| `midcurve-api` | 9 routes (incl. `/accounting`) | `serializers.ts`, `services.ts`, `discover`, `refresh-all` |
| `midcurve-business-logic` | rule + barrel + test | `rules/index.ts`, `workers/index.ts` |
| `midcurve-contracts` | 10 Solidity/deploy files | `package.json` |
| `docs` | 7 files | `architecture.md` |
| `midcurve-database` | — | `seed-contracts.ts` |

The UI and MCP server needed no edits — `VALID_PROTOCOL_VALUES` and all nine MCP `z.enum`s already listed only the two surviving protocols.

## Verification

| Suite | Before | After | Expected |
|---|---|---|---|
| `@midcurve/shared` | 127 | **127** | 127 ✓ |
| `@midcurve/api-shared` | 8 | **8** | 8 ✓ |
| `@midcurve/mcp-server` | 13 | **13** | 13 ✓ |
| `@midcurve/services` | 204 | **153** | 153 ✓ |
| `@midcurve/business-logic` | 36 | **24** | 24 ✓ |
| `@midcurve/onchain-data` | 7 | **7** | 7 ✓ |

- `pnpm build` — 11/11 tasks, clean
- `pnpm typecheck` — 14/14 tasks, clean. Note: typecheck was **already green on `main`**; the pre-existing `@midcurve/api` errors flagged as risk 4 in the plan no longer exist, so this is "stays green", not a delta report.
- `forge build --force` + `forge test` — 100 Solidity tests pass
- `git grep -i staking` over the tree returns only the eight files listed under *Deliberately left* below

## Database

**Production holds no staking positions.** That answer comes from the operator's direct knowledge — he never created one and is the only user of this single-tenant system — not from a query. Recording that here as the basis, rather than implying a database check happened.

Applied to the dev database, and needs the same on production:

```sql
DELETE FROM shared_contracts WHERE "sharedContractName" = 'UniswapV3StakingVaultFactory';
```

That was the only staking row anywhere: `positions`, `position_ledger_events` and `close_orders` all held zero. Sixteen surviving `shared_contracts` rows verified intact after the delete.

## Deploy step: orphaned RabbitMQ queue

`business-logic.uniswapv3-staking-post-journal-entries` (durable, bound to `positions.*.uniswapv3-staking`) survives in the broker. Unbind → check depth → delete, in the two phases settled on for #90.

Unlike #90 this queue **cannot fill**: once the position type is gone, nothing publishes that routing key again, so it sits empty rather than accumulating. Not urgent, but it shouldn't become permanent furniture.

## Deliberately left

- **`deployments/arbitrum.json`** keeps both staking entries — the contracts are deployed on Arbitrum and the file records what is on chain, not what the tree supports. `seed-contracts.ts` now logs `⊘ Skipping unknown contract` for them, which is the accurate outcome.
- **`docs/how-to-implement-new-positions.md`** still uses the staking vault as its running worked example. Rewriting it around the vault or retiring it is a separate decision, best made after #98 when the full staleness is visible.
- **The accounting layer**, per criterion 5: `TokenLotTransferEvent`'s `STAKING_DEPOSIT`/`STAKING_DISPOSE` in `types/accounting/token-lot.ts`, the Postgres enum values, and migration `20260504155248`. All four `STAKING_*` values *were* removed from `LedgerEventType` — different type, different file, shared spelling is not shared meaning.
- **Unrelated `'staking'` strings**: the `KnownProtocolAddress.interactionType` comments in `schema.prisma`, `transfer-classification.ts` and `known-protocol-address-input.ts`, plus CoinGecko token-list seed data.

## Note on the issue's premise

Acceptance criterion 4 was written on the assumption that an orphaned staking row would make `PositionFactory.fromDB` throw on every position-list load. It would not — the factory has one caller on a `uniswapv3`-only path, the list path hydrates nothing protocol-specific, and `parsePositionHash` already throws on staking hashes inside a `try/catch`. Confirmed and accepted in the thread; the row cleanup happened anyway, as hygiene.


0xNedAlbo:
## Review

Spot-checked the `discover` collapse on the branch, since that was the edit you flagged as most error-prone and where the fifth site turned up. It reads clean — the `Promise.all` is gone in favour of a direct await, no orphaned `stakingResult` or `allNewPositions`, and all four response fields read from `nftResult`. The doc comment matches what the endpoint now does.

The rest of the verification carries on its own: test counts landing exactly on the predicted per-file table, a clean grep sweep, `forge test` green, and CI in under three minutes. For a removal of this shape that is the right evidence.

Two things worth carrying forward rather than closing out here.

### The `package.json` script is a class, not a one-off

`deploy:staking-vault` pointed at a deleted file and survived build, typecheck, all six suites and `forge test`. Only the grep caught it. That is worth generalising before #98, which is a larger removal against a more entangled surface: **there is a set of files in this repo that no automated check reads.** Package scripts, `.env.example`, CI workflow step names, Prisma seed scripts, deployment configs, docs. A reference into deleted code from any of them is invisible to every gate we have.

The grep sweep should therefore be a required step in #98's verification recipe rather than a final courtesy, and it should explicitly cover those file types — not just `packages/*/src` and `apps/*/src`. Your §8 recipe already greps source; widen it.

### Manual deploy steps are accumulating in PR bodies

This PR carries two: the `shared_contracts` delete and the queue unbind/check/delete. That is the third time in this workstream — #78 had the row deletion, #90 had the v1 queue, now these. Each was correctly documented and each lives in a place nobody reads after the merge button is pressed.

A PR description is a record of intent, not a runbook. Once #99 merges, the only trace that two manual operations are outstanding is a paragraph in a closed pull request. #98 will add more of them — a data cleanup on `tokens`, an accounting schema drop, possibly more queues.

Worth deciding how these are tracked before #98 rather than after: a standing deploy-steps checklist issue, or a file in the repo that accumulates them and gets ticked off. Not for this PR. But two outstanding steps are manageable by memory and six are not.

### On risk 4

Reporting that the flagged typecheck errors no longer exist, and correcting your own note rather than quietly verifying against the weaker bar, is the right handling. Worth noting it cuts the other way too: `main` being green now means a typecheck regression in #98 will be unambiguous.

Approved to merge.

0xNedAlbo:
Note for the record, after merge: the two manual steps this PR carries — deleting the `shared_contracts` row for the staking vault factory, and unbinding then deleting the orphaned `business-logic.uniswapv3-staking-post-journal-entries` queue — **are not to be run against production.**

Jan's decision (2026-08-09): the current Railway environment is frozen at its deployed commit and receives no further deploys. At the end of the reduction the system is redeployed from scratch into a new project with an empty database, and the three live positions are re-imported. Both steps are therefore moot rather than outstanding — the environment they targeted is being replaced, and a fresh install seeds neither the row nor the queue.

Anyone finding this PR later and wondering whether something was left undone: it was not.

Issue #91, opened 2026-08-08
PR workflow enumerates test packages by hand, so new suites are invisible until someone notices


0xNedAlbo:
**Body rewritten 2026-08-12**, and **amended again the same day** after the membership work required by (5) contradicted part of the rewrite. The original framing — a hand-maintained enumeration of vitest packages that drifts — is correct but too narrow. Changes are marked below.

Split out of the #90 review, where the immediate instances were fixed and the mechanism was not.

## The problem

`.github/workflows/pr-tests.yml` names the packages it tests one step at a time:

```yaml
- name: Test @midcurve/shared
  run: pnpm --filter @midcurve/shared test:run
```

A hand-maintained enumeration drifts, and this one did: `@midcurve/services` (204 tests), `@midcurve/business-logic` (36) and `@midcurve/onchain-data` (7) were all outside the list. 247 tests that never ran against a pull request, for months, silently. Two of those suites predate this discovery by entire features — the business-logic tests landed with #78 and were green on that PR only because someone ran them locally.

#90 added the three missing steps. That fixes the instances. The next package to grow a suite will be invisible in exactly the same way, and nothing will say so — a test that never runs looks identical to a test that passes.

## What is missing today, beyond the enumeration

**`apps/midcurve-contracts` has a Solidity test suite that has never run in CI.** `test/treasury/` and `test/vault/` exist, and the package declares `"forge:test": "forge test"`. It is not in the workflow's enumeration, and — this is the point — **a workspace-wide `pnpm -r test:run` would not pick it up either**, because the script is not called `test:run`.

Those are the tests covering `MidcurveTreasury`, the contract from #124 that binds a single admin and operator and exposes `sweep()` / `rescueEth()`. They currently run only when someone remembers to run them locally.

The workflow already installs `foundry-rs/foundry-toolchain@v1` and already runs `forge build` in the treasury-artifact check, so the toolchain cost is already paid. Measured: +125 tests, no fork, no RPC, no secrets, 0.5s warm.

**`@midcurve/ui` has five tests that cannot pass.** They import `updatePositionInListCache`; that module was renamed to `invalidatePositionListCaches` in `6b49d484`. The test file was subsequently edited in `783e2295`, two months later, without the broken import being noticed — a file touched and never run. It is the cleanest instance of this issue's thesis in the repository.

[...]

Issue #95, opened 2026-08-09
Record the test baseline before the accounting rollback


0xNedAlbo:
Preparation for the accounting rollback. Investigation only — no code changes, no branch. The output is a comment on this issue.

## Why this is needed before anything is removed

The rollback will delete a large amount of code, and the test suites will shrink with it. Without a baseline recorded beforehand, a green run afterwards is uninterpretable: "247 tests, all passing" cannot be distinguished from "247 tests, of which 60 used to check something that is now gone and 12 quietly stopped being exercised".

This is cheap to establish now and impossible to reconstruct later.

## What to produce

A per-suite breakdown of which tests belong to the accounting layer and which do not. Accounting means journal entries, token lots, wallet perimeter, chart of accounts, and the double-entry posting rules in `midcurve-business-logic`.

For each of the two suites — `@midcurve/services` (204 tests) and `@midcurve/business-logic` (36) — report:

- Total test count and file count today, as the reference point.
- Which test files are wholly accounting: they exist only to exercise code that is being removed. These will disappear with the rollback.
- Which test files are mixed: they cover surviving behaviour but contain individual cases that touch accounting. Name the files and the approximate case counts. These are the risk — a mixed file loses cases silently during a rollback and the total still looks plausible.
- Which test files are unrelated. Count only; no need to enumerate.

Include `@midcurve/onchain-data` (7 tests) and the other three suites in the totals even if the expected answer is zero, so the numbers add up to a whole.

## How to report it

A table in a comment on this issue, plus the raw totals in a form that can be compared mechanically after the rollback. Prefer file-level granularity — that is the unit that will actually move.

If the boundary is unclear for a given file, say so rather than deciding. An ambiguous file is itself a finding: it means the accounting layer and the surviving core share a test, which tends to mean they share more than a test.

## Not in scope

Judging whether the tests are good, changing them, or proposing what the rollback should do. This issue establishes what exists.

0xNedAlbo:
# Test baseline before the accounting rollback

**Reference point:** `17c74e6a` (main, 2026-08-08T19:05:20+02:00)
**Measured by:** `pnpm --filter @midcurve/<pkg> test:run -- --reporter=json`, run for each of the six packages the PR workflow tests. All six were green at the time of measurement (395/395 passed, 0 failed, 0 skipped).

---

## 1. Totals — the reference point

| Suite | Test files | Tests | Wholly accounting | Mixed | Boundary (§5) | Unrelated |
[...]

PR #100, opened 2026-08-09
fix: resolve vault positions from lowercase addresses in the path

### LINKED ISSUE #80: Vault position routes 404 on lowercase addresses in the path
Split out of #77 (see the [decisions comment](https://github.com/0xNedAlbo/midcurve-finance/issues/77#issuecomment-5225632075), Q8).

## Problem

The vault routes build the position hash straight from the raw path parameters:

```ts
const positionHash = `uniswapv3-vault/${chainId}/${vaultAddress}/${ownerAddress}`;
```

The stored `positionHash` uses EIP-55 checksummed addresses, so a request carrying lowercase addresses — which is a perfectly valid way to write an EVM address, and what most tooling produces — resolves nothing and returns 404 "position not found".

## Affected routes

Everything under `apps/midcurve-api/src/app/api/v1/positions/uniswapv3-vault/[chainId]/[vaultAddress]/[ownerAddress]/` that builds the hash from unnormalized params, including `close-orders`, `close-orders/[closeOrderHash]/automation-state`, `ledger`, `accounting`, `apr`, `refresh`, `simulate`, and the position route itself. The `uniswapv3-staking` tree is worth checking for the same pattern.

## Fix

Normalize in the Zod path schema, as `close-orders/[closeOrderHash]/route.ts` now does (added in #78):

```ts
vaultAddress: z.string().regex(/^0x[0-9a-fA-F]{40}$/).transform((v) => normalizeAddress(v)),
```

Per `.claude/rules/evm-addresses.md`, use `normalizeAddress()` from `@midcurve/shared`, never viem's `getAddress()` directly.

0xNedAlbo:
Vault routes 404'd on a lowercase address in the path. The stored `positionHash` is EIP-55 checksummed; the routes built theirs straight from the raw path params, so a lowercase address — valid, and what most tooling emits — matched no row.

Closes #80

## Two layers, per the [decisions](https://github.com/0xNedAlbo/midcurve-finance/issues/80#issuecomment-5231005331)

**`UniswapV3VaultPosition.createHash()` (the real fix).** The format was a template literal at 17 sites — 14 routes, two service methods, one script. Only the service sites normalized, which is why the bug reproduced across a whole tree rather than in one file. The builder normalizes internally, so every construction is correct regardless of what the caller passes, including the call sites no Zod schema sits in front of. Fifteen copies of the format collapse to one — the change is a net reduction.

Modelled on `UniswapV3StakingPosition.createHash`, lifted from git history (`db9fe4b3^`) since #99 removed it, with normalization moved inside.

**The schema transform (second line of defence).** `GetUniswapV3VaultPositionParamsSchema` now normalizes both addresses. Every vault route parses through it, so all of them are fixed at the boundary too.

**Schema consolidation (Q1).** The three close-orders routes each had their own path schema — the copy #78 added, plus two that never normalized. All three now derive from the shared one. One definition of "a vault position path" instead of four.

## Verification

`pnpm typecheck` 14/14 · `pnpm build` 11/11 · all six CI suites green (shared 134, api-shared 22, mcp-server 13, services 153, business-logic 24, onchain-data 7). 21 of those tests are new.

Against the built artifacts, a lowercase, an uppercase-hex and a checksummed path all resolve to the same hash, and the builder normalizes unaided:

```
checksummed path -> uniswapv3-vault/42161/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/0x1F98431c8aD98523631AE4a59f267346ea31F984
lowercase   path -> uniswapv3-vault/42161/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/0x1F98431c8aD98523631AE4a59f267346ea31F984
uppercase   path -> uniswapv3-vault/42161/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/0x1F98431c8aD98523631AE4a59f267346ea31F984

all three resolve to one hash: true
builder normalizes unaided: true
```

**Not run: a live HTTP request against Arbitrum.** The mechanism above is verified against the compiled packages, not through a running server with auth and a database. Flagging rather than claiming it.

## Things worth stating rather than leaving implicit

**The EIP-55 checksum is not an integrity check here.** `normalizeAddress()` validates case-insensitively, so a mixed-case address with a *wrong* checksum is silently re-checksummed rather than rejected. That is consistent with the rest of the repo, and nothing in this system treats the checksum as a guard — but it means these endpoints accept an address a checksum-validating client would reject. Pinned by a test so the behaviour is deliberate rather than incidental.

**A throw inside a Zod `.transform()` surfaces as a 500, not a 400.** `normalizeAddress()` throws on invalid input. What keeps that unreachable is ordering — `.regex()` runs first and its pattern is strictly narrower than what `normalizeAddress()` accepts. It is ordering, not belt-and-braces, so it is now a comment next to the code.

**No DB check was run**, per the decisions: the environment is frozen and will be replaced by a fresh deploy with an empty database, so a lowercase hash row could not survive even if one existed. Both writers normalize in any case, and after this change they normalize through the builder.

**`chainId` gets stricter on three routes.** The shared parser rejects `0` and negatives where the local regexes took any digit string. A 400 on nonsense input that previously became a 404.

## Deliberately not done

- **`accounting/route.ts` is untouched** — it normalizes by hand, so it is not affected by this bug, and #98 removes it. Editing a file about to be deleted buys nothing. It is the one remaining inline hash literal in the routes.
- **No route-level HTTP tests**, and no vitest bootstrap for `apps/midcurve-api` — recorded on #91 instead, as agreed.
- **`discover` routes untouched** (Q3) — safe before via the service, safe now via the builder.
- **No redirect** (Q4) — normalize silently.
- **UI and MCP URL construction untouched** — those build request paths, not position hashes.
- **No separate audit issue** for the other address-carrying schemas in `@midcurve/api-shared` — centralising hash construction covers the half that matters.

### DIFF
--- a/apps/midcurve-api/src/app/api/v1/positions/uniswapv3-vault/[chainId]/[vaultAddress]/[ownerAddress]/apr/route.ts
+++ b/apps/midcurve-api/src/app/api/v1/positions/uniswapv3-vault/[chainId]/[vaultAddress]/[ownerAddress]/apr/route.ts
@@ -16,6 +16,7 @@ import {
   ErrorCodeToHttpStatus,
   GetUniswapV3VaultPositionParamsSchema,
 } from '@midcurve/api-shared';
+import { UniswapV3VaultPosition } from '@midcurve/shared';
 import type { AprPeriodsResponse, AprPeriodData, AprSummaryData } from '@midcurve/api-shared';
 import { serializeBigInt } from '@/lib/serializers';
 import { apiLogger, apiLog } from '@/lib/logger';
@@ -48,7 +49,7 @@ export async function GET(
       }
 
       const { chainId, vaultAddress, ownerAddress } = validation.data;
-      const positionHash = `uniswapv3-vault/${chainId}/${vaultAddress}/${ownerAddress}`;
+      const positionHash = UniswapV3VaultPosition.createHash(chainId, vaultAddress, ownerAddress);
 
       const dbPosition = await getUniswapV3VaultPositionService().findByPositionHash(user.id, positionHash);
 
--- a/apps/midcurve-api/src/app/api/v1/positions/uniswapv3-vault/[chainId]/[vaultAddress]/[ownerAddress]/close-orders/[closeOrderHash]/automation-state/route.ts
+++ b/apps/midcurve-api/src/app/api/v1/positions/uniswapv3-vault/[chainId]/[vaultAddress]/[ownerAddress]/close-orders/[closeOrderHash]/automation-state/route.ts
@@ -9,7 +9,6 @@
  */
 
 import { NextRequest, NextResponse } from 'next/server';
-import { z } from 'zod';
 import { withAuth } from '@/middleware/with-auth';
 import {
   createSuccessResponse,
@@ -18,7 +17,9 @@ import {
   ErrorCodeToHttpStatus,
   CloseOrderHashSchema,
   SetAutomationStateBodySchema,
+  GetUniswapV3VaultPositionParamsSchema,
 } from '@midcurve/api-shared';
+import { UniswapV3VaultPosition } from '@midcurve/shared';
 import { serializeCloseOrder } from '@/lib/serializers';
 import { apiLogger, apiLog } from '@/lib/logger';
 import {
@@ -30,10 +31,11 @@ import { createPreflightResponse } from '@/lib/cors';
 export const runtime = 'nodejs';
 export const dynamic = 'force-dynamic';
 
-const PathParamsSchema = z.object({
-  chainId: z.string().regex(/^\d+$/).transform(Number),
-  vaultAddress: z.string().regex(/^0x[0-9a-fA-F]{40}$/),
-  ownerAddress: z.string().regex(/^0x[0-9a-fA-F]{40}$/),
+/**
+ * The shared vault position path (which normalizes both addresses to EIP-55)
+ * plus the close order identifier.
+ */
+const PathParamsSchema = GetUniswapV3VaultPositionParamsSchema.extend({
   closeOrderHash: CloseOrderHashSchema,
 });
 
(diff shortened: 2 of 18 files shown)


0xNedAlbo:
## Review

The shape is right, and the doc comment in `get.ts` states it better than my instruction did: the schema is the outer guard, `createHash()` is the inner one for callers that never pass through a schema. That is the distinction that made Q2 worth inverting, and it now reads correctly to someone arriving cold.

Checked the diff. Twelve routes, both service call sites and the backfill script go through the builder; the two local schemas and the #78 copy collapse into the shared one; the transform ordering is documented next to the code rather than in a commit message. `createHash` throwing on an invalid address rather than emitting a malformed hash is the right choice and is pinned by a test.

**The count reconciles.** Post-#99 baseline is 332 across the six suites — 127 + 153 + 8 + 24 + 13 + 7. 353 is exactly that plus the 21 new. Worth saying explicitly, because that arithmetic working is what the #95 baseline was for.

Two notes, neither blocking.

**The last inline literal.** Leaving `accounting/route.ts` alone was correct — migrating a file #98 deletes is wasted motion, and it is not a latent instance of this bug since it normalizes by hand in the handler. But it means "no inline vault hash literal exists" is only true after #98. Worth a line in #98 so the last one is removed deliberately rather than incidentally; if #98 ever slips, someone should be able to find out that one was knowingly left.

**The missing live check.** Verifying against compiled packages rather than a running server with auth and a database is the honest boundary, and flagging it beats calling it verification. The right place for the live check is now the Phase 4 deployment guide: a lowercase path resolving a real position is a good smoke test for a fresh install, and it exercises the thing this PR fixes at the only moment someone will be making live requests deliberately anyway. Suggest it goes on that list rather than becoming a task nobody schedules.

The fourth instance on #91 is well placed — that a fix had nowhere in `apps/midcurve-api` to put its own regression test is a sharper argument than the enumeration drift on its own.

Approved to merge.

Issue #94, opened 2026-08-08
Debug logging is unreachable in development: pino.multistream drops every debug record regardless of LOG_LEVEL


0xNedAlbo:
**Rewritten 2026-08-13** against `4fff189` (merge of #147). The original filing described two observations and left the cause open. Both observations are real; they are **one defect**, not two, and the cause is established from the library source and reproduced. The title changed accordingly — `LOG_LEVEL` is not the mechanism.

**Amended 2026-08-13**, after the implementation plan's preflight. One claim of mine is withdrawn (see "A claim withdrawn"), and four decisions are added at the end of "Decisions taken". The preflight answers live in the plan comment; this body carries only what changes the specification.

## What was verified, and when

Verified 2026-08-13 by reading the code at `4fff189`, by reading the `pino` source at the version the lockfile resolves, and by reproducing the failure outside this repository.

Changes since filing (2026-08-08): none to the logging path. `logger.ts` was last touched by `98a8ee13` (2026-03-17). The close-order rule changed in #79, #88 and #86; the log statements named below were re-read at `4fff189` and the line numbers are current.

## The defect

`buildDestination()` in [`packages/midcurve-services/src/logging/logger.ts:85-88`](https://github.com/0xNedAlbo/midcurve-finance/blob/main/packages/midcurve-services/src/logging/logger.ts#L85-L88) builds the development destination as:

```
pino.multistream([{ stream: process.stdout }, { stream: fileStream }])
```

Neither entry carries a `level`. In `pino` 10.1.0 (the version `pnpm-lock.yaml` resolves for `midcurve-services`), `lib/multistream.js` `add()` assigns `level = DEFAULT_INFO_LEVEL` — 30 — to any entry that supplies neither `level` nor `levelVal`, and `write()` stops at the first stream whose level exceeds the record's. A debug record is level 20, so it reaches **neither** stream.

The logger's own level is irrelevant to this. `pino` filters twice: once at the logger, once per stream. Only the first is configured.

The same code is present in `pino` 9.14.0, the version the apps resolve.

### Reproduced

Outside the repository, against `pino` 10.1.0, with the exact `loggerConfig` from `logger.ts` and `level: 'debug'`:

| Destination | `logger.level` | `isLevelEnabled('debug')` | debug record written |
|---|---|---|---|
| A — `multistream([{stream: stdout}, {stream: file}])` (today's development branch) | `debug` | `true` | **no** |
[...]

Issue #82, opened 2026-08-08
Dead-lettered events are surfaced nowhere, and triggered close orders are discarded without a dead-letter path at all


0xNedAlbo:
**Rewritten 2026-08-14** against `c5487d4c`. The filing is accurate about the close-order DLQ and understates the situation in three ways: there are two dead-letter queues rather than one, the most consequential queue in the system has none at all, and a failure count that already exists is discarded before it would become visible.

Scope decided with Jan 2026-08-14: closing the silent-discard path belongs here, the numbers surface in the automation health endpoint, and the 7-day retention stays.

**Amended 2026-08-14** after the implementation plan. Two claims of mine are corrected (see "Corrections"), decisions 8–13 are added, criterion 4 is narrowed, criteria 9–10 are new, and the preflight questions are answered.

## What was verified, and when

Verified 2026-08-14 by reading the code at `c5487d4c`, by reading `amqplib` 0.10.9 source for the passive-declare behaviour the depth probe depends on, and — for the broker facts in "What must be established first" — from the live dev broker's management API, not inferred from source.

Changes since filing (2026-08-08): none to the dead-letter topology. #86 removed five swallowed reconciliation failures and #88 rebuilt the log sweep, but neither touched what happens to a message that cannot be processed.

### Two dead-letter queues, not one

| DLX | DLQ | TTL | fed by |
|---|---|---|---|
| `close-order-events-dlx` | `business-logic.uniswapv3-process-close-order-events.dlq` | 7 days | the close-order rule (from #78) |
| `domain-events-dlx` | `domain.events.dlq` | 7 days | every queue built by `setupConsumerQueue` (`packages/midcurve-services/src/events/topology.ts:295-302`), plus automation's `ORDER_EVENTS_QUEUE` (`uniswapv3-close-order-monitor.ts:585`) |

Neither has a consumer. Neither has a dead-letter exchange of its own. Both expire at 7 days. The filing describes the first; the second predates it and is in the same condition.

### The queue that matters most has no dead-letter path at all

`orders.pending` carries triggered close orders — the message that says a position reached its stop or target and should be closed. It is declared at `apps/midcurve-automation/src/mq/topology.ts:62-66` with `durable`, `exclusive: false`, `autoDelete: false`, and **no `x-dead-letter-exchange`**.

[...]

Issue #92, opened 2026-08-08
Five packages have a lint script and no ESLint config, so only the UI is actually linted


0xNedAlbo:
> **Status 2026-08-13.** PR #142 landed the lint gate and is merged. This issue is **not** finished: one piece of scope agreed in [this comment](https://github.com/0xNedAlbo/midcurve-finance/issues/92#issuecomment-5270716683) is still outstanding, deliberately kept here rather than opened as a separate issue. Reopened for that reason. See **Remaining work** immediately below; everything after it is the original lint scope, now done.

---

# Remaining work — PR 2: `prisma format`

`schema.prisma` is not in the format `prisma format` produces. Running it rewrites **59 lines**, so the next person to run the command — or any tooling that runs it — produces a diff that has nothing to do with their change. A repository-standard tool that nobody can run cleanly is the same defect class as the lint scripts this issue removed.

**Two commits, in this order, and the ordering is the point:**

1. `prisma format` and nothing else. No other file touched, so the reformatting is reviewable as a mechanical diff rather than hidden among changes.
2. The drift check plus its CI step.

**Implementation note established during planning:** `prisma format` has **no `--check` flag** in Prisma 6.19.0. The check must run the formatter and then use `git diff --exit-code`. Follow the treasury-artifact check already in `pr-tests.yml` — same structure, same error-message style — so the two read as one pattern rather than two inventions.

## What must be true afterwards

1. `prisma format` on a clean tree produces no diff.
2. A pull request that leaves `schema.prisma` unformatted fails CI, and the failure says what to run to fix it.
3. The check fails for formatting drift only, and does not depend on a database connection or on generated client output.

## Verification

Green CI after this lands proves only that the tree is currently formatted, not that the check would catch drift. Show it positively: introduce formatting drift into `schema.prisma`, show CI red with the intended message, revert, show green.

## Out of scope

Any change to the schema's *content*. This is a formatting-only change, and commit 1 must contain nothing else.

---

*Everything below is the original lint scope, delivered in PR #142.*

---

**Body rewritten 2026-08-12.** The original framing — five packages declare a `lint` script without an ESLint config, so only the UI is really linted — understates the problem and its table has gone stale. Verified against `main` at the time of writing:

- `.github/workflows/pr-tests.yml` has **no lint step at all**. Its seven steps are the test manifest check, the migration chain, build, typecheck, the Foundry artifact check, `forge test`, and the recursive unit tests.
[...]

PR #101, opened 2026-08-09
docs: record the one-tracker precondition in event to position resolution

### LINKED ISSUE #83: Document the single-tenant assumption in close order position resolution
Split out of the review on #78. Not a defect to fix — an assumption to write down, before someone discovers it by hitting the throw.

## The assumption

`UniswapV3ProcessCloseOrderEventsRule.findPositionForEvent()` resolves a vault close-order event to a position by chain, vault address and owner address. It does **not** scope by user, because an on-chain event carries no user.

`Position` is unique on `@@unique([userId, positionHash, ownerWallet])`, so two users who both track the same wallet legitimately produce two position rows for one `(chainId, vault, owner)` triple. `CloseOrder.positionId` is a single FK, so one order row cannot point at both. #78 therefore throws `UnroutableCloseOrderEventError` on more than one match rather than picking one arbitrarily — a state we cannot represent is not a state to guess at.

The consequence: on a multi-user install where two users track the same wallet, every close-order event for that vault dead-letters permanently, and re-dead-letters on every catch-up replay. The single-tenant assumption is now load-bearing in a place it was not before.

## Note on the NFT path

The NFT lookup has the same unscoped shape but resolves ambiguity differently: `positions.find(...)` takes the first match. Arguably worse — silent and arbitrary rather than loud — but it is pre-existing behaviour and #78 deliberately did not change it.

## What this issue is asking for

Documentation, in the place a future reader will look:

- A note in `.claude/rules/platform-agnostic-design.md` (or wherever the multi-tenancy stance belongs) stating that on-chain event → position resolution is single-tenant by assumption.
- A pointer from the two lookups to that note.

If the project ever does go multi-tenant, this issue is the trailhead: the fix is not in the lookup but in the data model — an order row would need to fan out per position, or close orders would need to hang off the wallet rather than the position.

0xNedAlbo:
Closes #83

Documentation only. No behaviour change, no migration, no new tests.
Implements the decisions in [this comment](https://github.com/0xNedAlbo/midcurve-finance/issues/83#issuecomment-5231093072).

## The note

New section in `.claude/rules/platform-agnostic-design.md`, next to the existing Identity section (D1). It states the narrow, mechanical claim (D2) — *at most one user tracks any given on-chain position* — not "the product is single-tenant", which would contradict deliberately multi-user code elsewhere and get disbelieved on the first counterexample.

It also records the NFT lookup's silent first-match as a known deviation rather than filing it (D5): same precondition, two failure modes, deliberately not unified.

## Pointers

| Location | Why there |
|---|---|
| `findPositionForEvent()` JSDoc | The lookups the note is about |
| The NFT `positions.find()` at the first-match | Names the deviation at the line that performs it |
| The `UnroutableCloseOrderEventError` string (D3) | The only surface anyone encounters when this fires — #94 established debug logging can't be turned on, #82 that nothing surfaces dead letters |
| `CloseOrder.orderIdentityHash` (D4) | Where the constraint is **enforced**, not merely surfaced |

## The schema comment correction

`Position.positionHash` carried:

```
// This allows multiple users to import/watch the same on-chain position
```

That promise does not hold end to end. The allowance is real at the `Position` table — per-user import dedupe, `@@unique([userId, positionHash, ownerWallet])` — and downstream the same state permanently dead-letters that position's close-order events. The comment now says what is actually true at each layer and points at the note.

Not a claim of single tenancy, and no code was removed to match it. Per the issue discussion, removing the aspirational half means unthreading `userId`, which belongs to the reduction phase and not to this PR.

## Verification

- `pnpm typecheck` — 14/14 tasks pass
- business-logic suite — 24/24 pass, unchanged. The ambiguity test asserts `/Ambiguous vault position/` against the unchanged prefix, so the extended error string does not affect it
- `prisma validate` — schema valid

Note on the schema file: `prisma format` reformats ~100 unrelated lines across the file. I reverted that churn and applied the comments by hand, so the diff is 19 lines in one file rather than 115.


0xNedAlbo:
## Review

The note does what it needed to. It states the narrow mechanical precondition rather than a product stance, it names the disagreement between the Position layer and the close-order layer instead of describing both neutrally, and it records the NFT first-match as a deviation with its reason rather than quietly unifying it. "Not a representable state" is the right framing — the constraint is in the schema, not in a policy someone chose.

**One gap, in the section immediately above yours.** The *Identity* list gives `uniswapv3/{chainId}/{nftId}` for positions and stops there. Vault positions have a second format, and #100 just made `UniswapV3VaultPosition.createHash()` its single source of truth. A reader now hits an identity list that omits the format, then a section three lines later that discusses `(chainId, vaultAddress, ownerAddress)` resolution in detail.

Add the vault line and point it at the builder. Two lines, same file you already have open, and it closes the loop between #100 and this note — the document that explains why vault identity includes the owner should also be the document that says what the hash looks like.

**`prisma format` — your call was right, and it should be recorded rather than left as a surprise.** The ~100 lines are pre-existing drift, not churn this PR introduced, so keeping the diff at 19 lines is correct: a documentation change should not carry a hundred lines of unrelated reformatting, and mixing them would make both harder to review.

But "the schema is not in `prisma format` shape and running the formatter produces a large unrelated diff" is a trap of exactly the kind this repo keeps yielding — the next person to run it will assume they broke something, or will land the churn inside an unrelated PR where nobody can see past it. Add it as a comment on #92, which already holds the lint-config equivalent. Same family: a tool the repo nominally uses that nobody can run cleanly.

**Branch naming — the repo's history is the stronger signal.** `type/issue-NN-slug` is what the last several branches used, and `docs/` as the type is right for this one.

Worth flagging separately, though: if a skill or command file in this repo still prescribes `feat/spec-NNNN-`, that is a leftover of the SPEC-NNNN regime, which was retired. It should not be quietly worked around by each person who hits it. Say which file it is and I will decide whether it gets cleaned up now or filed.

Approved to merge once the Identity lines are in.

0xNedAlbo:
## Review addressed

**Identity list — added** in `657396da`. The vault format is `uniswapv3-vault/{chainId}/{vaultAddress}/{ownerAddress}`, both addresses EIP-55 checksummed, pointed at `UniswapV3VaultPosition.createHash()` as the source of truth per #100. It also notes *why* the owner is in the identity — a vault is an ERC-20, so the holder is part of it — which is the fact the section below then relies on.

One small thing beyond the two lines: I relabelled the existing entry from `Positions` to `Positions (NFT)`. With two position formats in the list, an unqualified `Positions` line reads as the format for all positions, which is the ambiguity that caused the gap. Revert it if you would rather keep the diff at exactly two lines.

**`prisma format` — recorded on #92**: https://github.com/0xNedAlbo/midcurve-finance/issues/92#issuecomment-5231119880

Includes the reproduction, the confirmation that the drift is non-semantic (`prisma validate` passes either way), and the sequencing constraint on any fix — the formatting-only commit has to land alone or it defeats its own purpose.

**Branch naming — the file is [`.claude/commands/sync-issue-feedback.md:18`](.claude/commands/sync-issue-feedback.md#L18).**

```
  `feat/spec-NNNN-<short-slug>` (use the SPEC number from the issue
   title). Implement in logical commits, push, open a PR linked to
```

It is the only occurrence — `grep -rn "spec-NNNN\|SPEC" .claude/commands/` returns that one line across all ten command files, so the rest of the SPEC-NNNN regime is already gone from them. Worth noting the instruction is doubly stale: it prescribes both a retired numbering scheme *and* a fixed `feat/` type, which is wrong for this PR and for the `fix/` and `refactor/` branches in recent history. The repo convention it should describe is `{conventional-commit-type}/issue-NN-<slug>`.

I have not touched it — your call whether it gets cleaned up now or filed.

Verification unchanged and re-run on the new commit: `pnpm typecheck` 14/14, business-logic 24/24, `prisma validate` clean.

Not merging — over to you.

Issue #102, opened 2026-08-09
Retired SPEC-NNNN regime still prescribed in .claude/commands/


0xNedAlbo:
The SPEC-NNNN way of working was retired. Residue of it survives in `.claude/commands/`, where it is unfollowable and gets improvised around rather than reported.

## The instance that surfaced it

`.claude/commands/sync-issue-feedback.md:18`, under *Explicit go-ahead to implement*:

> create a feature branch `feat/spec-NNNN-<short-slug>` (use the SPEC number from the issue title)

Two dead references in one instruction. There are no SPEC numbers in issue titles any more, so the parenthetical cannot be satisfied, and the branch pattern does not match anything in the repository's history. Recent branches use `<type>/issue-<NN>-<slug>` — `fix/issue-80-vault-hash-normalization`, `docs/issue-83-single-tracker-assumption`, `refactor/issue-97-remove-uniswapv3-staking`.

Replace it with the convention that is already in use:

> create a feature branch `<type>/issue-<NN>-<short-slug>`, where `<type>` is the conventional-commit prefix for the work (`feat`, `fix`, `refactor`, `docs`, `chore`).

## The part that matters more

Fixing the one line found by accident is not the point. An instruction that cannot be followed should have been reported the first time someone hit it, and instead it was quietly worked around — which means there is no way to know how much else in that directory is in the same state.

Sweep all of `.claude/commands/` and `.claude/rules/` for further residue of the retired way of working:

- other SPEC-NNNN references, or instructions to derive anything from a SPEC number
- doubt-pass, reconciliation-pass or tier vocabulary
- ADR ceremony
- RFC numbering or authoring instructions
- the unified-diff-patch convention for document amendments
- references to files, directories or workflows that no longer exist

**Report the list before changing anything beyond the branch line.** Some of it may be worth keeping despite its origin, and that is Jan's call rather than a cleanup decision. For each finding, say what it instructs, whether it is currently followable, and whether anything in the repository still matches it.

## Not in scope

`CLAUDE.md` itself, unless the sweep turns up a contradiction with it — in which case report rather than resolve.

Anything about how the commands *should* work. This issue is about statements that are false, not about improving the workflow.

0xNedAlbo:
## Sweep report — `.claude/commands/` and `.claude/rules/`

Nothing changed yet. Every file in both directories read; findings below with the three facts this issue asks for per item: what it instructs, whether it is currently followable, and whether anything in the repository still matches it.

[...]

Issue #93, opened 2026-08-08
Dead delivery paths in the close-order subsystem: WS provider and non-finalized catch-up are exported and never called


0xNedAlbo:
**Body rewritten 2026-08-13.** The findings are unchanged and still true against `main`. What has changed is that the open question in the original body — whether we want a non-finalized catch-up at all — is answerable from the code, and is answered here rather than left to the implementation. A stop condition has been added to the sweep.

Surfaced while planning the AC 8 verification in #87 (findings F5 and the note on the non-finalized catch-up). Deliberately kept out of that pass: a verification run must not carry code changes, or the measurement stops meaning anything.

## What was found — re-verified 2026-08-13

**`UniswapV3CloserSubscriptionBatch`** — `apps/midcurve-onchain-data/src/ws/providers/uniswap-v3-closer.ts`. A WebSocket subscription batch plus a factory. `WorkerManager` constructs and starts six subscribers, and the close-order one is `CloseOrderSubscriber` from `workers/uniswapv3/uniswapv3-close-order-poller.ts` — the polling path. Nothing in `workers/index.ts` references the WS closer.

**`executeCloseOrderCatchUpNonFinalized`** and `…ForChains` — `apps/midcurve-onchain-data/src/catchup/close-order-catchup.ts`, re-exported from `catchup/index.ts`. `CloseOrderSubscriber.runCatchUp()` calls `executeCloseOrderCatchUpFinalizedForChains` and nothing else.

Both read as capability. Neither runs. #79's original premise was wrong for exactly this reason — it assumed the non-finalized catch-up republished by design, and it does not, because it is never called.

## Decision: finalized-only is what we ship

**Delete the non-finalized catch-up.** The reasoning is in the code and should be restated in the PR so this is not re-litigated later:

1. **The non-finalized range is already covered — by the poller.** `UniswapV3CloserPollingBatch.poll()` scans `lastProcessedBlock + 1` → `client.getBlockNumber()`. That is the chain head, not the finalized head. Catch-up leaves the cursor at the finalized block, the poller picks it up and covers the remainder on its next cycle.

2. **The only thing a non-finalized catch-up would buy is latency**, bounded by the poll interval. `CLOSER_POLL_INTERVAL_MS` defaults to `3600000` — one hour. That is the entire benefit: closing a window of up to an hour at startup.

[...]

Issue #87, opened 2026-08-08
Verify AC 8 of #77: fallback poller picks up a directly-registered vault close order


0xNedAlbo:
## Context

#77 was closed as completed with acceptance criterion 8 unverified:

> The fallback poller picks up a vault order registered by direct contract interaction, without the UI involved.

The poller never ran during the verification session — catch-up was still grinding when work stopped, and catch-up blocks poller startup. The gap is recorded in a comment on #77, but #77 is closed, so that comment appears in no list and no board. This issue exists so the outstanding verification is visible rather than archived.

This is the failure shape #77 itself was about: something real, correctly noted, and invisible to anyone not already looking at the right closed thread.

## What is already known to work

Step A passed on-chain during the #77 verification: registering a stop-loss on a vault position produced exactly one `CloseOrder` row with the correct position link, an EIP-55 identity hash, shares stored as a string, an active monitor subscription, and an empty DLQ. The reconciliation writer is therefore known good. What is untested is the *other* writer — the fallback poller — and its behaviour against a row reconciliation already created.

## Precondition

**Corrected 2026-08-08.** An earlier revision of this issue said `CATCHUP_BATCH_SIZE_BLOCKS` must be set or catch-up would not complete. That was wrong: `getCatchUpConfig()` in `apps/midcurve-onchain-data/src/lib/config.ts` defaults it to `10000`, so the variable is always set and catch-up always completes. Nothing is missing from the configuration.

The problem is the *value*, not its absence. At `10000` the startup backfill measured 7m43s on Base alone and had not finished on Arbitrum after 8 minutes, and no poller starts on any chain until every chain's backfill completes — see #88. Verification at the default will therefore stall the same way the original session did, without any misconfiguration to find.

Set `CATCHUP_BATCH_SIZE_BLOCKS` high for this run (the Alchemy key is on a paid plan, where the range is unlimited on Ethereum, Arbitrum and Base). `.env.example` documents the variable and the caveat.

## What to verify

1. Restart `onchain-data` with a raised batch window and confirm catch-up completes and the poller starts within one cycle.
2. The poller picks up the vault close order that already exists in the database, registered by direct contract interaction — no UI involvement.
3. The upsert path is exercised against the row reconciliation wrote: the poller must update rather than insert a second row.
4. After the poll cycle, there is still exactly one `CloseOrder` row for that position and trigger mode, and no duplicate entry in the automation log.

## What counts as evidence

[...]

Issue #61, opened 2026-05-02
SPEC-0003a: Staking Vault for Uniswap V3 Positions — Implementation Spec


0xNedAlbo:
# SPEC-0003a: Staking Vault for Uniswap V3 Positions — Implementation Spec

**Date:** 2026-05-02
**Status:** ready for implementation
**Source:** Derived from local RFC-0003. The RFC process is in transition;
for now the coding spec is posted on GitHub while the RFC remains local.
**Audience:** Coding agent (Claude Code).

This spec describes *what* to implement. Conceptual rationale and
alternatives live in the RFC. Read the spec for code; consult the RFC
only if a concrete decision in the spec seems ambiguous.

This revision incorporates three extensions over the prior local version:

- **Top-up.** `stake()` is callable in `Empty` (initial mint) AND `Staked`
  (additive `increaseLiquidity`).
- **Partial unstake.** A `pendingBps` slot in `[0, 10000]` parametrises
  `swap()` and `quoteSwap()`. When `pendingBps == 0`, the default behaviour
  is full close (effective bps = 10000), preserving the prior executor-
  initiated full-close path.
- **Fractional flashClose.** `flashClose(bps)` accepts a bps parameter for
  partial owner-initiated close.

The Settled-state reward slots (`baseReward`, `quoteReward`,
`principalUnstaked`, `rewardsClaimed`) are replaced by a buffer pattern
that allows `unstake()` and `claimRewards()` to be called repeatedly across
the vault lifecycle.

---

## 1. Architecture

- **`StakingVaultFactory`** — singleton contract per chain. Deploys
  `StakingVault` clones via EIP-1167 minimal proxy.
- **`StakingVault`** — single implementation contract. Each clone wraps
  exactly one Uniswap V3 position and is bound to exactly one owner.
- **Non-transferable ownership.** The owner address is set once during
  clone initialization. Once set, it MUST NOT be mutable. No transfer
  function, no ownership token, no upgrade mechanism that could rebind it.
- **Atomic create-and-initialize.** The factory's `createVault()` MUST
  perform the clone deployment AND initialization atomically in the same
  call frame. This closes the standard EIP-1167 race-condition where a
  third party could front-run an external `initialize()` call. The
  implementation's `initialize()` SHOULD additionally revert on a second
  invocation (`initializer` modifier from OZ Initializable, or equivalent
  one-shot guard).
- **One vault per UV3 position.** A new clone is deployed per
[...]

PR #104, opened 2026-08-09
refactor: drop CloseOrder.sharedContractId and the retired SPEC command residue

### LINKED ISSUE #81: Close orders created by the event rule have no sharedContractId, those created by refresh() do
Split out of #77 (see the [decisions comment](https://github.com/0xNedAlbo/midcurve-finance/issues/77#issuecomment-5225632075), Q7). Recorded as an observation; deliberately left alone in #78 so NFT behaviour did not change as a side effect.

## Observation

`CloseOrder.sharedContractId` is a nullable FK to the closer contract in the shared-contract registry. The two writers of these rows disagree about it:

- `UniswapV3CloseOrderService.refresh()` looks the contract up anyway (it needs the address to call `getOrder`) and passes `sharedContractId` on create.
- `UniswapV3ProcessCloseOrderEventsRule` takes the contract address from the event envelope and never resolves it to a registry row, so it creates orders with `sharedContractId: null`.

This holds for both `uniswapv3` and `uniswapv3-vault` — it is not a vault-specific gap.

## Consequence

Whether a close order is linked to its `SharedContract` row depends on which writer happened to create it, which is a race, not a property of the order. `@@index([sharedContractId])` suggests the link is meant to be queryable; today such a query silently misses every event-created row.

## Options

1. Resolve the contract in the rule (one registry lookup per registration) and set the field.
2. Drop the field and derive the contract from `config.contractAddress` where needed.

Either is fine; the current state — the same field meaning two different things depending on provenance — is the part worth removing.

0xNedAlbo:
Three commits, two issues. Bundled into one PR at Jan's request; the branch is named for the code half.

## 1. `docs(claude)` — retired SPEC regime and MCP transport references (#102)

`sync-issue-feedback.md` told you to branch as `feat/spec-NNNN-<short-slug>` and take the number from the issue title. No open issue carries one; the only two that ever did, #61 and #63, are closed and describe the staking vault deleted in #99. Replaced with `<type>/issue-<NN>-<short-slug>`, the convention the branch history already uses.

Three commands routed their work "via the github MCP", which is configured in no `.mcp.json` and in no session. Rewritten to name the outcome rather than the transport, so they survive the next change of environment.

**One addition beyond the approved scope, flagged for veto:** `research-plan.md:13` said "Read issue #$1 using the GitHub tools" — the same defect in weaker wording, in the same directory, which the sweep under-reported. Folded in rather than left to be rediscovered. Say the word and I will drop it.

Also fenced `bug-fix.md`'s commit template, which rendered as one paragraph and lost the blank line before the trailer.

Left alone as agreed: Finding C, and all four items outside the swept directories.

## 2. `fix(services)` — normalize contractAddress (#81)

The precondition for commit 3, and it turned up while verifying Jan's condition rather than being planned.

`CloseOrder.config.contractAddress` held a checksummed address when `refresh()` created the row and a lowercase one when the event rule did — the same coin flip as `sharedContractId`, one level down, in the field commit 3 makes load-bearing. All three producers funnel through `buildCloseOrderEvent` and all three lowercase first, so normalizing in the decoder fixes them together, the way #100 put normalization inside `createHash` rather than at each call site.

`vaultAddress` is deliberately **not** normalized here: it feeds `buildCloseOrderRoutingKey()`, so changing its case would change routing keys and break existing queue bindings. The rule already normalizes it on the way into config.

Three tests added, including one pinning that the routing key does not move.

Side effect worth naming: the WS provider's `|| 'unknown'` fallback now throws instead of writing a literal `'unknown'` into a field used as a contract address. That line is untouched — it lives in `UniswapV3CloserSubscriptionBatch`, exported and never called (#93), and dies with that deletion.

## 3. `refactor(services,database)` — drop the field (#81)

Set by `refresh()`, null from the event rule, so its meaning depended on which writer won the race.

Removed rather than populated, because **nothing reads it**: `closeOrder.sharedContract` is never included or selected anywhere, so `@@index([sharedContractId])` backed no query. The index was the strongest argument for keeping the link and nothing ever cashed it in. Meanwhile the executor already reads `config.contractAddress` and calls the contract at it — the derivation is the existing mechanism, with the FK sitting beside it as decoration.

Populating it in the rule would also have encoded something false: `refresh()` resolves the *latest* closer for the chain, while an order lives at whatever address its event carried. Those diverge on the next redeploy.

The FK's one real consumer was #87, which used null-vs-set to tell the writers apart while verifying the fallback poller. Closed.

### No data migration

Existing rows keep whatever their writer produced. The environment they live in is frozen and gets replaced by a fresh deploy with an empty database, so converting them would be work with no consumer. Stating it rather than leaving it as an unasked question.

### The migration is hand-written

`prisma migrate dev` refuses to run here: the dev database has `20260529000000_init` applied with no local directory, so it demands a full reset. That drift is pre-existing and unrelated to this change — `prisma migrate status` reports the schema up to date, which is why it has gone unnoticed. The migration SQL is `prisma migrate diff` output and applies cleanly on a database without the drift. **Worth its own issue.**

## Verification

- `pnpm typecheck` — 14/14 tasks green
- `@midcurve/services` — 156 tests pass
- `@midcurve/business-logic` — 24 pass, including the 24-test close-order event rule suite
- `@midcurve/onchain-data` — 7 pass

## Also settled, no code

`CloseOrder.config.nftId` is a string while `Position.config.nftId` is a number. Exactly one site compares them — `uniswapv3-process-close-order-events.ts:551-553` — and it converts explicitly with `parseInt`, so the behaviour is correct and the item is cosmetic. Recorded, closed, unchanged.

Closes #81
Refs #102, #93

### DIFF
--- a/.claude/commands/bug-fix.md
+++ b/.claude/commands/bug-fix.md
@@ -32,8 +32,11 @@ This workflow is asynchronous, including when Claude Code runs in Plan Mode. Whe
 
 6. **Commit**: Create a commit with a descriptive message:
 
-fix: short description
-Fixes #<number>
+   ```
+   fix: short description
+
+   Fixes #<number>
+   ```
 
 7. **Open a PR**: Push the branch and create a pull request:
    - Title: `fix: short description`
--- a/.claude/commands/finish-issue.md
+++ b/.claude/commands/finish-issue.md
@@ -3,8 +3,8 @@ The PR for issue #$ARGUMENTS has been merged into main.
 1. Switch to `main` and `git pull`.
 2. Delete the feature branch locally (`git branch -d <name>`) and
    on the remote (`git push origin --delete <name>`).
-3. Close issue #$ARGUMENTS via the github MCP with a short comment:
-   2–3 lines summarising what was actually delivered, plus a link
-   to the merge commit or PR.
+3. Close issue #$ARGUMENTS with a short comment: 2–3 lines
+   summarising what was actually delivered, plus a link to the
+   merge commit or PR.
 
 Do not touch any other branches or issues.
--- a/.claude/commands/plan-issue.md
+++ b/.claude/commands/plan-issue.md
@@ -1,4 +1,4 @@
-Read GitHub issue #$ARGUMENTS using the github MCP.
+Read GitHub issue #$ARGUMENTS.
 
 Plan the implementation thoroughly. Cover:
 - What needs to change (files to add/modify, key types/functions, tests).
--- a/.claude/commands/research-plan.md
+++ b/.claude/commands/research-plan.md
@@ -10,7 +10,7 @@ You are about to work on a RESEARCH issue. Before doing the research, post a pla
 ## Task
 
 1. Determine the current repository (e.g. via `git remote -v`).
-2. Read issue #$1 using the GitHub tools.
+2. Read issue #$1.
 3. Understand scope, out-of-scope, and the desired output format from the issue body.
 4. Post a plan comment with the following structure:
 
--- a/.claude/commands/sync-issue-feedback.md
+++ b/.claude/commands/sync-issue-feedback.md
@@ -1,4 +1,4 @@
-Sync with issue #$ARGUMENTS via the github MCP.
+Sync with issue #$ARGUMENTS.
 
 1. Fetch the issue and any linked PR.
 2. Read all comments newer than your last activity on the issue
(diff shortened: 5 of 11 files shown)


0xNedAlbo:
## Review

Read the migration and the schema change rather than trusting the suites — green tests say nothing about a hand-written migration.

The three commits are in the right order and the second genuinely earns the third. The comment left in `schema.prisma` where the field used to be is the best part of this PR: it records *why* the column went, so the next person to want a contract reference on a close order finds the reasoning instead of rediscovering it. Same for the decoder comment explaining why `contractAddress` is normalized and `vaultAddress` deliberately is not — the routing-key consequence is exactly the thing that would otherwise be broken by a well-meaning later edit.

**The scope overrun is accepted.** `research-plan.md:13` is the same defect in weaker wording, and folding it in beats leaving it for rediscovery.

The miss is worth more attention than the edit, though. The sweep's value was its claim to completeness — five categories reported empty, Finding A naming three files — and a fourth instance existed in a paraphrase. A sweep keyed on the exact phrase "github MCP" cannot find "the GitHub tools". That generalises directly to #98, where I made a grep sweep a required verification step: keyed on symbol names and concepts it will work, keyed on phrasings it will miss the same way. Worth carrying into that plan.

**One nit on the migration.** `DROP INDEX "close_orders_sharedContractId_idx"` carries no schema qualifier while the `ALTER TABLE` below it says `"public"."close_orders"`. It will work as long as `search_path` includes `public`, which it does — but this is the one artifact in the PR that cannot be verified locally, in a database with two schemas, and it should not depend on session state to succeed. Qualify it.

**The Prisma drift needs its own issue, and I have filed it** — see below. Writing the migration by hand rather than resetting Jan's database was the right call, and reporting that `prisma migrate status` says "up to date" while `migrate dev` demands a reset is the finding, not a footnote.

**On the `|| 'unknown'` line:** normalizing at the decoder defusing it without touching it is a better outcome than editing dead code. It now throws instead of propagating a literal into a field the executor calls contracts at, and the line still dies with #93.

**#70 is already closed** — I closed it from my side after your case-1 answer, so `/finish-issue` has nothing to do there.

Approved to merge once the index drop is qualified.

Issue #103, opened 2026-08-09
Deploy the reduced system from scratch on Railway, and write the deployment guide while doing it


0xNedAlbo:
The final phase of the platform reduction, and the point of all of it. Jan's decision (2026-08-09): rather than migrating the existing environment forward, the whole system is deployed fresh — new Railway project, empty database, everything from scratch — and the deployment guide is written *during* that pass rather than afterwards from memory.

**This happens last**, after the accounting rollback (#98) and the bug phase. Deploying an unfinished state and then migrating it again by hand is exactly what this avoids.

## Why this shape

Midcurve holds almost no authoritative state of its own. Positions, ledger events, cost basis, PnL, APR periods and close orders are all projections of on-chain data and are reconstructed from the chain. What the database holds beyond that is a handful of inputs: the wallet list, favourite pools, column preferences. Rebuilding from empty is therefore the natural deployment mode for this system rather than a last resort.

The consequence is that every manual deploy step accumulated during the reduction becomes moot rather than outstanding — the row deletions in #78 and #99, the queue operations in #90 and #99, the data cleanup and schema drop in #98. The environment they targeted is being replaced. Each is recorded in its own PR as not-to-be-run.

## Sequencing constraint

**The current Railway environment is frozen** at its deployed commit and receives no further deploys until this phase. It keeps running, and keeps monitoring Jan's three live positions and their SL/TP orders, until the new environment is up and the positions have been re-imported. Only then is the old project decommissioned. There must be no window in which nobody is watching the triggers.

## What this phase produces

1. A new Railway project with an empty database, running the reduced system.
2. Jan's three positions re-imported and verified, including their close orders.
3. A deployment guide in the repository, written as the deploy happens.

## Prerequisites collected during the reduction

These are known before the phase starts and should be handled as part of it rather than rediscovered:

- **The seed must not warn about things that are fine.** `apps/midcurve-contracts/deployments/arbitrum.json` still lists `UniswapV3StakingVault` and `UniswapV3StakingVaultFactory`, deliberately kept as a true record of what is deployed on-chain (#97). `seed-contracts.ts` has no kebab mapping for either, so every seed run prints two `⊘ Skipping unknown contract` lines. The fix belongs in the seed, not in the record. This matters here specifically: the first fresh seed is the run where someone establishes what normal output looks like, and expected warnings must not be present at that moment.
[...]

PR #106, opened 2026-08-09
chore(database): verify the migration chain and close the drift that hid it

### LINKED ISSUE #105: Prisma migration history has drifted from the dev database, and migrate status reports it as healthy
Surfaced while implementing #81 (PR #104), where writing a nine-line migration required working around the tooling entirely.

## What is broken

The dev database has migration `20260529000000_init` recorded as applied, and no corresponding directory exists under `prisma/migrations/`. The recorded history and the checked-in history have diverged.

Consequences observed:

- **`prisma migrate dev` refuses to run.** It sees a database whose history it cannot reconcile with the folder and demands a reset of both schemas — which would destroy the working dev database. Every migration therefore has to be hand-written from `prisma migrate diff` output.
- **`prisma migrate status` reports "Database schema is up to date!"** That is a false all-clear. The one command whose job is to tell you the state of migrations reports health while the workflow underneath it is unusable.

## Why it matters more than it looks

**#98 is the large one.** The accounting rollback drops six models, an enum, an entire Postgres schema and seven back-relation fields. That migration will also have to be hand-written and cannot be validated by applying it locally in the normal way. A nine-line hand-written migration is reviewable; that one is not, in the same sense.

**#103 depends on the chain being correct.** The fresh deploy applies the full migration history to an empty database and must land on exactly the schema the code expects. Nobody has done that. If the folder and the real schema have drifted — and the recorded history says they have — then the first time anyone finds out is during the deploy that this whole reduction is aiming at.

That is the concrete risk: the migration chain has never been verified to reproduce the current schema from empty, and the phase that assumes it will is the last one.

## What would resolve it

Not prescribing the fix. The shape of the question:

- Establish whether applying the checked-in migrations to an empty database produces the schema the Prisma client expects. This is cheap to test against a throwaway database and answers the #103 risk directly.
- Decide what to do about the divergence: baseline the dev database against the folder, reconstruct the missing migration, or squash the history to a single initial migration. #103 already notes squashing as a question for the fresh-deploy phase, when an empty target makes it safe.
- Make `migrate status` meaningful, or record clearly that it is not to be trusted here. A command that reports health during a broken state is worse than one that errors.

## Sequencing

The verification step — does the chain reproduce the schema from empty — is worth doing **before #98**, not after. It is a throwaway database and an afternoon, and it determines whether the largest migration of the reduction is being written against a foundation that works.

The cleanup of the divergence itself can wait for #103.

0xNedAlbo:
Closes #105

The migration chain was verified clean before any of this was written — 34 migrations apply to an empty database and produce `schema.prisma` exactly. The divergence was one foreign row and one unapplied migration, not a broken history. This PR turns that one-off verification into a command, records the two tooling facts that made the state hard to read, and closes the process gap that let dev drift in the first place.

## Database work (already done, not in the diff)

Both operations were approved on the issue and have been applied to `midcurve_dev`:

**1. The phantom row, captured before deletion:**

```
id                  | 28095b4b-6bd2-4245-9280-0e53661e292b
checksum            | 0233d93cd369fecf70015353388709996565a935a2d6aace2c8d43f39224c1fc
migration_name      | 20260529000000_init
started_at          | 2026-05-31 08:55:18.814956+00
finished_at         | 2026-05-31 08:55:18.842766+00
applied_steps_count | 1
rolled_back_at      | (null)
logs                | (null)
```

`DELETE FROM _prisma_migrations WHERE migration_name = '20260529000000_init';` → `DELETE 1`

**2. `pnpm db:migrate:deploy`** → applied `20260809132600_drop_close_order_shared_contract_id`, the migration from #81 / PR #104 that had reached `main` but never been applied.

**Left alone deliberately:** the `evm` schema (`block`, `chain_state`, `transfer_event`, 0 rows). It belongs to `midcurve-agent`, not to Midcurve, and it is the only remaining trace of the contamination. It disappears on its own when #103 creates a fresh database.

**State afterwards:**

```
$ npx prisma migrate status
34 migrations found in prisma/migrations
Database schema is up to date!

$ npx prisma migrate diff --from-schema-datasource … --to-schema-datamodel …
-- This is an empty migration.
```

The single `close_orders` row survived and is still `monitoring`; only the dead `sharedContractId` column went, as #81 decided.

## `db:migrate:verify`

The obvious one-liner would have been a script that never works. Two findings forced a real one:

- **Prisma does not create the shadow database.** `--shadow-database-url` pointing at a non-existent database fails with `P1003`. It has to be created and dropped around the diff.
- **A `SHADOW_DATABASE_URL` in `.env` is invisible to an npm script.** The shell expands CLI flags before Prisma loads `.env`, so the flag would have received an empty string. The URL is derived from `DATABASE_URL` instead, so there is nothing to configure and nothing to forget.

Both paths are tested, including the one that matters:

```
$ pnpm db:migrate:verify                    # green
✓ The migration chain reproduces schema.prisma exactly.        exit 0

$ pnpm db:migrate:verify                    # red (temporary field added to schema.prisma)
[*] Changed the `shared_contracts` table
  [+] Added column `driftProbe`
✗ Drift: applying the migration chain to an empty database does not
  produce schema.prisma. A fresh deploy would land on the wrong schema.
                                                                exit 2
```

The first version of the script printed the drift and exited **0** — `$?` after a failed `if` with no `else` is the status of the `if` statement, not of the command. Caught by running the red case; that is why the red case is in this description.

## Stated trigger

Per the condition on the issue, the script is documented as manual with an explicit trigger rather than added silently. `CLAUDE.md` carries it next to the command that creates migrations — the point where someone is already about to need it — and `docs/architecture.md` states when to run it: after creating a migration, before applying the chain to any empty database, and after rebasing onto a `main` that gained migrations.

It is **not** wired into `pr-tests.yml`. That needs a Postgres service container, and per Q3 the recommendation was manual now and revisit at #103. Say the word and I'll add the service — it is about six lines.

## The two tooling facts

Recorded in `docs/architecture.md`, because both cost real time to establish and neither is discoverable from the CLI:

- **`migrate status` does not report database-only migration rows** unless a local migration also happens to be unapplied. Reproduced on a freshly created, provably correct database, so it is a property of the tool and it follows onto the fresh Railway instance. Also posted to #103.
- **Such a row cannot be removed with `migrate resolve`** — `--rolled-back` fails `P3012`, `--applied` is for the opposite case. The direct `DELETE` is the supported route, documented as such rather than left looking like a shortcut.

## The process gap

`.claude/commands/finish-issue.md` gains a step: after pulling `main`, check whether the pull brought migrations and apply them. Nothing in the workflow required a merged migration to be applied anywhere, which is exactly how #104's migration reached `main` and stopped there.

## Verification

`pnpm typecheck` — 14/14 tasks successful. No test changes: the diff is one shell script, one npm script and three documentation files, and the verification *is* the test.

### DIFF
--- a/.claude/commands/finish-issue.md
+++ b/.claude/commands/finish-issue.md
@@ -1,9 +1,17 @@
 The PR for issue #$ARGUMENTS has been merged into main.
 
 1. Switch to `main` and `git pull`.
-2. Delete the feature branch locally (`git branch -d <name>`) and
+2. If the pull brought new migrations —
+   `git diff --name-only ORIG_HEAD HEAD -- packages/midcurve-database/prisma/migrations/`
+   — apply them to the dev database:
+   `cd packages/midcurve-database && pnpm db:migrate:deploy`.
+   Merging a migration does not apply it anywhere. Dev drifts silently
+   after every schema change until someone runs `migrate dev` and
+   inherits the divergence. Say what was applied, or that there was
+   nothing to apply.
+3. Delete the feature branch locally (`git branch -d <name>`) and
    on the remote (`git push origin --delete <name>`).
-3. Close issue #$ARGUMENTS with a short comment: 2–3 lines
+4. Close issue #$ARGUMENTS with a short comment: 2–3 lines
    summarising what was actually delivered, plus a link to the
    merge commit or PR.
 
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -21,6 +21,23 @@ jobs:
       contents: read
       pull-requests: read
 
+    # Throwaway Postgres for the migration-chain verification below. Holds no
+    # data and is never seeded — the check creates and drops its own database.
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -38,6 +55,15 @@ jobs:
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
+
+      # Does applying the checked-in migrations to an empty database produce the
+      # schema the client expects? Hand-written migrations are standing practice
+      # here, so this is the expected failure of the workflow rather than a rare
+      # one. Runs before the build for fast feedback; it needs only the schema.
+      - name: Verify migration chain
+        run: pnpm --filter @midcurve/database db:migrate:verify
+        env:
+          DATABASE_URL: postgresql://postgres:postgres@localhost:5432/postgres
 
       - name: Build
         run: pnpm build
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,8 +41,8 @@ pnpm db:migrate:verify  # Does the chain reproduce schema.prisma from empty?
 pnpm db:studio          # Inspect database
 ```
 
-Run `db:migrate:verify` after creating a migration and before applying the chain
-to any empty database. `prisma migrate status` does not report database-only
+`db:migrate:verify` runs on every PR; run it locally after writing a migration
+to get the answer sooner. `prisma migrate status` does not report database-only
 migration rows, so a clean status is not evidence of a clean history — see
 [docs/architecture.md](docs/architecture.md), "Verifying the migration chain".
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -37,8 +37,14 @@ cd apps/midcurve-contracts && forge build  # Solidity contracts
 ```bash
 cd packages/midcurve-database
 pnpm db:migrate:dev --name migration_name
-pnpm db:studio  # Inspect database
+pnpm db:migrate:verify  # Does the chain reproduce schema.prisma from empty?
+pnpm db:studio          # Inspect database
 ```
+
+Run `db:migrate:verify` after creating a migration and before applying the chain
+to any empty database. `prisma migrate status` does not report database-only
+migration rows, so a clean status is not evidence of a clean history — see
+[docs/architecture.md](docs/architecture.md), "Verifying the migration chain".
 
 ## Architecture Docs
 For detailed architecture, auth flows, and design decisions:
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -768,12 +768,56 @@ cd packages/midcurve-database
 # Edit prisma/schema.prisma
 pnpm db:migrate:dev --name your_migration_name
 
+# Verify the chain still reproduces the schema from empty
+pnpm db:migrate:verify
+
 # Changes immediately available via workspace linking
 # All apps automatically use the updated client
 
 # Run Prisma Studio to inspect database
 pnpm db:studio
 ```
+
+#### Verifying the migration chain
+
+`pnpm db:migrate:verify` applies every checked-in migration to a throwaway
+database and diffs the result against `schema.prisma`, exiting non-zero if they
+differ. It answers the only question that matters about the migration folder:
+**does applying it to an empty database produce the schema the client expects?**
+
+Run it:
+
+- after creating a migration, especially a hand-written one
+- before applying the chain to any empty database (a fresh deploy, a new dev
+  machine, CI)
+- after rebasing a branch onto a `main` that gained migrations
+
+It needs `psql` and a `DATABASE_URL` whose role may `CREATE DATABASE`. The
+throwaway database is created and dropped by the script; `DATABASE_URL` itself
+is only read.
+
+#### Two things `prisma migrate` will not tell you
+
+**`migrate status` does not report database-only migrations.** If
+`_prisma_migrations` holds a row with no corresponding folder, `migrate status`
+prints `Database schema is up to date!` and says nothing — *unless* a local
+migration also happens to be unapplied, at which point it reports both. A clean
+`migrate status` is therefore not evidence of a clean history. This reproduces
+on a freshly created, provably correct database, so it is a property of the
+tool, not a symptom of a damaged one. `db:migrate:verify` is the check that
+does not have this blind spot. (Found in #105.)
+
+**A database-only migration row cannot be removed with `migrate resolve`.**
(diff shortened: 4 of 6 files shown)


0xNedAlbo:
## Review

**The script that exited 0 on drift is the finding here, and it is worth stating plainly.** A check that reports success when it has detected a failure is worse than no check, because it converts an absent guard into a false assurance — and you only found it by running the red case.

That is the fifth instance of one pattern in this workstream. The CI filter that listed packages it never ran (#91). Five `lint` scripts with no config (#92). The heartbeat that recorded the chain head as scanned blocks (#89). `migrate status` reporting health over a phantom row (#105). Every one of them looked like a working check, and every one was only exposed when someone constructed the failure it was supposed to catch. Verifying that a check *fails when it should* is not extra diligence; it is the only part of writing a check that establishes anything.

The `P1003` and env-var findings vindicate the condition on their own. A script requiring a `SHADOW_DATABASE_URL` that an npm invocation cannot see would have been broken on first run and quietly abandoned.

**On the trigger: put it in CI, and I will take the six lines.**

Documented-manual with three stated occasions is precisely the shape this repository has been failing at all week. But the deciding argument is not consistency, it is the failure mode: this check detects `schema.prisma` disagreeing with the migration chain, and hand-written migrations are now the *standing practice* here rather than an exception. #104's was hand-written. #98's will be, and it drops six models, an enum and an entire schema.

So the drift this catches is not rare — it is the expected failure of the current workflow, arriving on the largest migration of the reduction. A manual check is at its least likely to be run exactly when it matters most.

Adding it also removes the caveat you recorded on #103 about #98 having to re-run the verification: in CI it re-runs by itself, and nobody has to remember.

**`.env.example` deviation approved.** A documented-but-unused environment variable is the same defect in miniature, and deriving the URL removes the configuration rather than documenting it.

**The caveats you added to the #103 record are the right ones.** Point-in-time and structure-not-seeds are both real limits, and a verification that reads as broader than it is would be its own kind of false assurance.

**One question before merge:** if the verification fails partway, does the throwaway database get dropped? A check that leaves a stray database behind on the red path is a small thing until it happens three times and someone starts wondering which one is real. If there is no cleanup on failure, add it.

0xNedAlbo:
Both addressed. Two commits: `9ae8b1d` (cleanup) and `005465b` (CI).

## Your question: yes, but I only half knew it

I had assumed the `trap … EXIT` covered it. Rather than answer from the code, I constructed both failure paths:

| path | exit | throwaway DB |
|---|---|---|
| drift detected | non-zero | dropped |
| SIGTERM mid-run (verified alive when signalled, success line never printed) | 143 | dropped |

So the answer to what you asked was already yes. What was missing was everything around it, now in `9ae8b1d`:

- **Explicit `INT` and `TERM` traps.** A signal does not reliably run the `EXIT` trap, so Ctrl-C and a CI cancellation are handled directly, exiting 130 and 143. The EXIT trap happened to fire under `SIGTERM` in my test; that is not a guarantee worth relying on.
- **A failed `DROP` now says so** instead of being silenced. The old cleanup redirected its own stderr to `/dev/null` — so the one case where a stray database really would be left behind was also the one case that printed nothing. That is the same shape as the bug you're describing: a cleanup that reports success when it failed.
- **`.env` is sourced only when `DATABASE_URL` is unset**, so an explicit value wins over the file. Needed for CI, and the previous order was surprising anyway.

One honest gap: **`SIGKILL` still leaves the database behind**, and no trap can catch it. The script drops the database *before* creating it, so a leak from a hard kill is cleaned up by the next run rather than accumulating. That is mitigation, not prevention, and it is documented as such.

## CI: done, and I checked it can fail

`pr-tests.yml` gets a Postgres service and one step, on the existing job so there is no second checkout and install. It runs before the build — it needs only the schema, so failure arrives in about a minute rather than five.

Green run on this PR: the step executes in **1.4s** and reports `No difference detected.`

But a green check proves nothing about a check, which is the whole content of your review. So I opened a throwaway PR (#107, now closed, branch deleted) with a single field added to `schema.prisma` that no migration creates:

```
[*] Changed the `shared_contracts` table
  [+] Added column `driftProbe`
✗ Drift: applying the migration chain to an empty database does not
  produce schema.prisma. A fresh deploy would land on the wrong schema.
##[error]Process completed with exit code 2.
```

Failed step: `Verify migration chain`. Only that step. Exit 2, run red in 1m3s.

That is the fifth instance you describe, applied to the thing I had just built: I had added a CI check and confirmed only that it passes. If it had been miswired — wrong `DATABASE_URL`, `psql` absent, exit code swallowed by `pnpm --filter` — the green run would have looked exactly the same.

I also removed the `#98` caveat from the [#103 record](https://github.com/0xNedAlbo/midcurve-finance/issues/103#issuecomment-5231803401) accordingly — the verification is no longer something anyone has to remember to re-run. The other two caveats stand: it is still point-in-time in the sense that it verifies the chain as it exists on each PR, and it still proves structure rather than seeds.

## Docs updated to match

`CLAUDE.md` and `docs/architecture.md` said "run it manually, here are three occasions". They now say it runs on every PR, with the reason stated as you put it: hand-written migrations are standing practice, so this is the expected failure of the workflow rather than a rare one. The cleanup guarantee and the `SIGKILL` gap are recorded there too.

CI on this branch is green. Nothing merged.

0xNedAlbo:
## Review

Both items are answered better than they were asked.

**The stray-database question.** You could have answered it from the code and been right. Constructing both paths instead turned up the thing reading would have missed: the cleanup silencing its own stderr, so the single case where a database really would be left behind was the case that printed nothing. That is the pattern in miniature, found inside the fix for the pattern.

`SIGKILL` leaking, with the self-heal on next run documented as mitigation rather than prevention, is the honest end state. A trap that cannot exist should not be implied to exist.

**PR #107 is the part I want to single out.** I wrote that verifying a check fails when it should is the only part of writing a check that establishes anything, and you applied it to your own work in the same session — a throwaway field, a red run, one named failing step. Without it, a miswired check would have produced a green run indistinguishable from a correct one, and we would have carried a CI step into #98 believing it guarded the largest migration of the reduction.

That is now the standard for this workstream: a check arrives with its red case, or it arrives unverified.

**The #103 edit.** Correcting the record rather than appending to it is right, and marking the edit inline keeps the trail honest. The two remaining caveats are the true ones.

**On the job name — leave it, and it is yours to sequence, Jan.** "Build, typecheck & unit tests" now also verifies migrations, so the label is stale. But the name is *configuration* rather than documentation: branch-protection rules reference it by string, so renaming it silently would disable the protection it names. Not touching it unilaterally was correct.

If it gets renamed, do it in one motion with the protection rule, and pick something that will not need renaming again the next time a step is added.

Approved to merge. Remember the two database steps are already applied to dev and are not part of the merge.

PR #107, opened 2026-08-09
DO NOT MERGE: prove the migration-chain check fails

0xNedAlbo:
Throwaway PR. Adds a field to schema.prisma that no migration creates, to confirm the CI check added in #106 actually goes red. Closing immediately after.


0xNedAlbo:
Confirmed what it was opened for: the migration-chain check fails on drift (exit 2, `Verify migration chain` step). Discarding.

Issue #96, opened 2026-08-09
Dependency inventory of the accounting layer, and the cut line against position-ledger


0xNedAlbo:
Preparation for the accounting rollback. Investigation only — no code changes, no branch. The output is a comment on this issue, and it becomes the basis for the rollback issue itself.

## Context

The platform is being distilled to a reduced core: UniswapV3 position management, tokenized vault positions, the workflow from pool search to opening a position, and SL/TP close orders. The accounting layer is being rolled back — journal entries, token lots, wallet perimeter, chart of accounts, and the double-entry posting rules in `midcurve-business-logic`.

Nobody has yet established how far that layer reaches. The rollback cannot be scoped until someone has.

## What to produce

An inventory of everything that would have to move, organised by layer. For each item, state what it is and whether it is wholly accounting, partly accounting, or merely a consumer.

- **Prisma schema** — models, enums, relations and indexes belonging to accounting. Note especially any FK from a surviving model into an accounting one; those are the places where removal stops being additive.
- **Services** in `@midcurve/services` — `journal`, `token-lot`, `wallet-perimeter` are the obvious ones. Report what else calls into them.
- **Rules** in `midcurve-business-logic` — the three `PostJournalEntries` rules and anything else bound to the same exchanges. Report which routing patterns become unbound.
- **API routes** — the `accounting` subtrees under both position protocols, plus anything else that reads accounting data.
- **UI** — tabs, panels, hooks and queries that render it. Note whether removal leaves a hole in a surviving screen or removes a whole screen.
- **MCP tools**, if any expose accounting.
- **Environment variables, seeds, scripts** that exist only for it.

## The question that matters most

**Where is the cut line against `position-ledger`?**

The ledger is not on the removal list. The close-order executor depends on it, PnL and APR are computed from it, and it is the producer whose events the journal rules consume. So the expected answer is that the ledger stays and the consumers go.

What needs establishing is whether that line is actually clean:

- Does anything in the ledger exist *only* to serve accounting — event types, fields, computed values that no surviving consumer reads?
- Does any accounting code write to the ledger, rather than only reading from it?
- Are there ledger event types that exist solely to be posted as journal entries?

Report what you find. The decision is Jan's; this issue supplies the material for it.

## How to report it

A comment on this issue, organised by the layers above. Where something is partly accounting, say which part — that is where the rollback will need judgement rather than deletion.

[...]

PR #108, opened 2026-08-09
refactor: roll back the accounting layer

### LINKED ISSUE #98: Roll back the accounting layer

0xNedAlbo:
Removes the accounting layer: journal entries, token lots, the chart of accounts, and the double-entry posting rules. 89 files, +329 / −8,093.

The `position-ledger` stays, and so do PnL, cost basis, APR and the ledger UI — all ledger-owned, per §8 of the [inventory](https://github.com/0xNedAlbo/midcurve-finance/issues/96#issuecomment-5230757456).

Closes #98

## Four things worth saying plainly

**1. `GET /user/me` never returned `reportingCurrency`.** Only `PATCH` did. So `get_user` has been emitting `reportingCurrency: null` for as long as it has existed — a field that was silently null in a tool response, which is a small instance of the thing this rollback is about. Acceptance criterion 7's GET clause was already satisfied before this branch.

**2. The slimmed wallet-change rule is NFT-only by fact, not by omission.** `UniswapV3ReevaluateOnWalletChangeRule` kept its `recalculateAggregates` call and lost everything else, including the entire vault branch. That branch never called `recalculateAggregates` and could not usefully have done so — `UniswapV3VaultLedgerService.recalculateAggregates(isToken0Quote, tx?)` takes no wallet set, because a vault position is tied to a single `ownerAddress` and has no per-event `isIgnored`. Everything the branch did do was journal and lot maintenance. The missing vault branch is the accurate description of what this rule does today.

Class name, rule name, file name and queue are unchanged; only the directory moved, since `rules/accounting/` is gone. Nothing about the queue or its bindings changed, so it is reused rather than replaced — deliberately not the #90 situation.

**3. The migration's `DELETE` is outside the verification.** `db:migrate:verify` compares structure, so a hand-appended data statement is checked by eye and nothing else. That is acceptable for one narrow DELETE and it should be visible rather than implied.

It turned out to be worse than that, in a way worth recording. `prisma migrate diff` generated **only** the `users` column drop and the `known_protocol_addresses` table drop. Once `schema.prisma` stops declaring `schemas = ["public", "accounting"]`, Prisma does not inspect the accounting schema at all, so it reports no drift whether the schema is dropped or not — and `db:migrate:verify` passes either way. **Three of the four statements in the migration are hand-written**, and the schema drop is one of them.

Statement order is load-bearing: `accounting.token_lots.tokenId` references `public.tokens(id)` with no `onDelete`, i.e. `RESTRICT`, so the schema must go before the token rows.

**4. No deploy notes.** Frozen environment, fresh Railway deploy later, per the scope change on the issue.

## What the symbol sweep found that nothing else would have

Run by symbol before any code was written — ~60 identifiers, whole tree, not just `src`.

**`apps/midcurve-api/Dockerfile:38` and `apps/midcurve-api/entrypoint.sh:16`.** The chart of accounts had a *third* seeding path, alongside `db:seed:accounts` and the business-logic worker startup: the API image bundles `seed-accounts.ts` with esbuild at build time and the entrypoint runs the result on **every container start**. Deleting the seed and stopping there would have broken the Docker build at the esbuild step, and past that, the container at boot — after `pnpm build`, `pnpm typecheck` and all six suites had gone green.

## Verification

| Suite | Predicted | Actual |
|---|---|---|
| `@midcurve/services` | 156 | **156** |
| `@midcurve/shared` | 134 | **134** |
| `@midcurve/business-logic` | 24 | **24** |
| `@midcurve/api-shared` | 22 | **22** |
| `@midcurve/mcp-server` | 13 | **13** |
| `@midcurve/onchain-data` | 7 | **7** |

Predicted [before implementing](https://github.com/0xNedAlbo/midcurve-finance/issues/98#issuecomment-5231971960). `pnpm build` and `pnpm typecheck` green (14/14 tasks). `db:migrate:verify`: "The migration chain reproduces schema.prisma exactly."

**One file changed and did not change count.** `user-settings-service.test.ts` had a single `costBasisMethod` assertion (the prediction said two — the other four references were fixture keys). It proved that an absent key is filled from `DEFAULT_USER_SETTINGS`, so it is retargeted at `poolTableVisibleColumns`, which reaches the same code path. 26 in, 26 out.

**Criterion 6, against a copy of the dev database.** Before: 14 `erc721` token rows, 6 token lots, accounting schema present, `reportingCurrency` column present. After applying the chain: 0 rows of any uninstantiable type, schema gone, column gone, `known_protocol_addresses` gone — and positions (6) and ledger events (31) unchanged.

**Criterion 2, against a second copy.** Exercised the exact path the slimmed rule keeps — build the wallet set via `UserWalletService`, then `UniswapV3LedgerService.recalculateAggregates(...)` — on all three UniswapV3 NFT positions in the dev data:

| Position | events | ignored with real wallet set | ignored with empty set |
|---|---|---|---|
| `cmnvusa48…` | 11 | 6 | 11 |
| `cmnzrmrxo…` | 6 | 3 | 6 |
| `cmnzrptl9…` | 4 | 1 | 4 |

Recalculating with the real set is a no-op against the current state (so the rule is idempotent and the data is already correct), an empty set flips every event to ignored, and restoring the real set restores the previous values exactly. Nothing is posted, because there is nowhere left to post to.

## What this deliberately does not do

- **`UserWallet` and `UserWalletService` are untouched** — the four `UniswapV3PositionService` call sites, the SIWE login path, the wallets API and `WalletManagementPage` all stand. "Remove the wallet perimeter" taken literally would have broken position import. Only the half with no callers went: `EvmWalletPerimeterService`, `KnownProtocolAddressService`, `TransferClassification`, and the `KnownProtocolAddress` model (0 rows). The `wallet-perimeter/` directory keeps its name — the perimeter check is surviving behaviour.
- **The wallet domain events stay bound.** `wallet.added` / `wallet.removed`, `ROUTING_PATTERNS.WALLET_*`, `WalletChangedPayload`, `buildWalletRoutingKey` — all still have a consumer.
- **`position-ledger`, APR, PnL, close orders and pools are untouched.**
- **`basic-currency` is untouched** (3 rows in dev) — dead on its own account, belongs to #93.
- **No migration squashing** — that is the fresh-deploy pass.
- **No new tests.** Adding one would move counts the acceptance criteria rest on. Criterion 2 is verified manually above instead.
- **No `MISSION.md` change** — held outside the repository.

## Two notes for elsewhere

- The `@midcurve/ui` test-script gap is [filed on #91](https://github.com/0xNedAlbo/midcurve-finance/issues/91#issuecomment-5231956559), not fixed here — adding a script would add 5 cases to the totals above.
- `fix(ui): fall back to the overview tab for an unrecognised tab param` is commit 1 and stands on its own. Checking it properly showed it is not rollback fallout: both detail pages read `?tab=` and render a flat list of `activeTab === … &&` guards with no default, so **any** unrecognised value renders an empty page today. `?tab=accounting` bookmarks are just the instance this branch would otherwise have created.


0xNedAlbo:
## Review

Read the migration rather than trusting the green checks — it is the one artifact here that no automation covers.

It is right. `DROP SCHEMA … CASCADE` before the token cleanup, because `accounting.token_lots.tokenId` restricts; the `DELETE` schema-qualified; `IF EXISTS` making it idempotent on a fresh target. And the comment block explains the ordering rather than just asserting it, which is the difference between a hand-written migration someone can review and one they have to reverse-engineer.

**The Prisma blind spot is the sixth instance of the pattern this week, and the most consequential.**

`db:migrate:verify` derives its scope from the same declaration the migration changes. Take a namespace out of `schemas = […]` and it leaves the check's field of view — not the database. A migration that omitted the schema drop would have verified clean, passed CI, and left the entire accounting layer standing on every fresh deploy. The check we built last week specifically to catch this class would have reported the chain reproduces the schema exactly.

That is not a defect in the check; it is a structural limit of comparing a declaration against a database. **Removals cannot be verified by a tool whose scope comes from the declaration.** It generalises past Prisma.

**One condition before merge: record it where it will be found.** The PR body is not a durable home — we established that with the deploy steps, and this is a fact with a much longer half-life. It belongs in `docs/architecture.md` alongside the two other hard-won Prisma facts from #105, and on #103, because the fresh deploy is exactly where a silently-surviving schema would land and where someone will be trusting `db:migrate:verify` to tell them the database is right.

**The `entrypoint.sh` find is the strongest argument for the sweep instruction.** A third seeding path, running on every API container start, past build, typecheck and all six suites. On a fresh Railway deploy that is a container that boots and immediately fails against a schema that no longer exists — found by grep, in a file no check reads.

**Six of six predictions exact.** Third time the #95 instrument has paid off, and the first time on a change of this size. The self-correction on the assertion count belongs in that column too: one, not two, and retargeting it at `poolTableVisibleColumns` keeps the code path rather than the count.

**Criterion 2 is verified the right way.** Real wallet set is a no-op, empty set flips every event, restoring returns the exact prior values. That design tests idempotence, effect and reversibility with three runs against real data — better than any assertion I would have asked for.

**One observation for #103, not for here.** A fresh deploy replays the whole chain, so it will create the entire accounting layer in migration `20260227144408` and tear it down in this one. Harmless, and the concrete argument for squashing when the target is empty.

Approved to merge once the blind spot is recorded in `architecture.md` and on #103.

0xNedAlbo:
## Condition met — `04479eed`

Recorded in both places.

**`docs/architecture.md`** — the section head goes from "Two things `prisma migrate` will not tell you" to three, with the new fact first. It states the mechanism (the check reads its scope from the same declaration a removal changes), the concrete instance, and the generalisation: *whatever the datamodel stops declaring, the diff stops seeing*. Data statements are folded in under the same argument, since `migrate diff` compares structure and a `DELETE` is invisible to it for the same reason a dropped schema is. It ends with what to do instead — write the drop by hand, apply the chain to a copy of a real database, query for what should be gone, and say so in the PR, because the green check does not mean what it usually means.

**#103** — https://github.com/0xNedAlbo/midcurve-finance/issues/103#issuecomment-5232168689. Same fact, framed for the moment it bites: "the chain verifies" is evidence that everything the datamodel declares is present, not that everything it no longer declares is absent. With three post-migration queries to put in the deployment guide, since that closes the gap the automation structurally cannot.

I also added it to the root `CLAUDE.md`, next to the existing `migrate status` warning — four sentences. That was not in your condition; it seemed like the highest-traffic place the fact belongs, given that the existing sibling fact is already there. Say the word and I will drop it.

### On your #103 observation

Folded into that comment rather than left as a remark, because it is a stronger argument than "harmless". The fresh deploy will create the accounting schema in `20260227144408`, refactor it, extend it twice, then drop all of it in `20260809162923` — plus the same round trip for staking from #97. It nets to the right schema, which is exactly what `db:migrate:verify` confirms. But it means the migration folder describes two subsystems that no longer exist, and anyone reading it to understand the schema reads both before reaching the live one. That is the case for squashing while the target is empty: not tidiness, but that the folder stops describing a system that is gone.

### One correction to the review

> `IF EXISTS` making it idempotent on a fresh target

Slight overstatement on my part in how I wrote it, worth being precise about since you read the migration closely. `DROP SCHEMA IF EXISTS` is there for the case where the chain is applied to a database that never had the schema — but on a genuine fresh install the chain *does* create `accounting` in `20260227144408`, so the schema is always present when this migration runs. The `IF EXISTS` is belt-and-braces, not load-bearing. The `DELETE` is the one that is genuinely a no-op on a fresh target, and deliberately so.

Not merging, per your instruction. Ready when you are.

0xNedAlbo:
## Correction accepted, and both extensions stand

**You are right about `IF EXISTS` and I was wrong.** On a genuine fresh install the chain creates `accounting` in `20260227144408`, so the schema is always present when this migration runs — the guard is belt-and-braces, not load-bearing. It matters only for a target where something else already removed the schema, which is not a case the chain produces. The statement that is actually a no-op on an empty target is the `DELETE`, deliberately so. Worth having straight, since the next person to read that migration will be trying to work out which parts are structural.

**`CLAUDE.md` — keep it.** Beyond the condition, and the right call: it is the highest-traffic location, the sibling fact already lives there, and a fact separated from its sibling is half as likely to be found by someone who needs both.

**The #103 argument is better than my version.** I offered the create-then-destroy as an observation with squashing attached as tidiness. Your framing is the substantive one: the migrations folder now narrates two subsystems that do not exist, and anyone reading it to understand the schema reads both before reaching the live one. That is a claim about what the folder *communicates*, not about how neat it is — and it survives the objection that the chain works, because nobody disputed that it works.

The three post-migration queries for the deployment guide are a good addition. They turn "the accounting layer is gone" from something the deployer believes into something they check.

Approved to merge.

PR #109, opened 2026-08-09
fix(ui): disable the transaction prompt button when the action is unavailable

### LINKED ISSUE #42: Bug: Collect button remains enabled when wallet is on wrong network
## Description

In the Collect Fees dialog for a Uniswap V3 (vault) position, when the connected wallet is on a different chain than the position, the dialog correctly shows a "Wrong network. Please switch to {chainName}" warning with a "Switch" button — but the primary "Collect" button next to it is **not disabled**. It uses the same blue primary styling, `disabled: false`, `aria-disabled: null`, `pointer-events: auto`, full opacity, and is fully clickable. A user can click it and the execute path is only gated programmatically, which produces a confusing failure rather than a clear UI signal.

Observed in the dashboard Collect Fees modal for a PAXG/USDC position on Ethereum while the wallet was connected to Arbitrum (and again when connected to Sepolia).

## Expected Behavior

While the "Wrong network" warning is shown, the "Collect" button should be visibly disabled:
- `disabled` attribute set
- reduced opacity / `cursor-not-allowed`
- no hover state change

Only the "Switch" button should be actionable. Once the wallet is on the correct chain, the warning disappears and "Collect" becomes enabled normally.

## Root Cause

Lives in the shared `useEvmTransactionPrompt` helper, not in the Collect Fees form itself. The form already passes the right signal:

`apps/midcurve-ui/src/components/positions/protocol/uniswapv3/uniswapv3-collect-fees-form.tsx` (lines 116–138):

\`\`\`tsx
const canCollect =
  isConnected &&
  !isWrongNetwork &&
  !isWrongAccount &&
  unclaimedFees > 0n;

const collectFeesTx = useEvmTransactionPrompt({
  // ...
  enabled: canCollect,     // ← correctly false on wrong network
  onExecute: () => collectFees.collect(),
});
\`\`\`

`canCollect` correctly becomes \`false\` on wrong network, and is passed as \`enabled\`. \`handleExecute\` inside the hook does gate on \`enabled\` (line 252: \`if (!enabled) return;\`), so clicking is a no-op — but the **button element never reflects that state visually**:

`apps/midcurve-ui/src/components/common/EvmTransactionPrompt.tsx` (lines 267, 319–324):

\`\`\`tsx
const showButtons = showActionButton && (isIdle || isError) && !isActive;
// ...
{showButtons && (
  <button
    onClick={isError ? handleRetry : handleExecute}
    className=\"px-4 py-1.5 bg-blue-600 text-white text-sm rounded-lg font-medium hover:bg-blue-700 transition-colors cursor-pointer\"
  >
    {isError ? retryButtonLabel : buttonLabel}
  </button>
)}
\`\`\`

The button has no \`disabled={!enabled}\`, no disabled styling, and \`enabled\` is not part of \`showButtons\`. So in the idle state the button renders fully active regardless of \`enabled\`.

## Affected Files

- `apps/midcurve-ui/src/components/common/EvmTransactionPrompt.tsx` — primary fix location (button at lines 319–324). Add \`disabled={!enabled}\` and disabled styling (e.g. \`disabled:opacity-50 disabled:cursor-not-allowed disabled:hover:bg-blue-600\`).
- `apps/midcurve-ui/src/components/positions/protocol/uniswapv3/uniswapv3-collect-fees-form.tsx` — downstream consumer, no change needed.
- All other \`useEvmTransactionPrompt\` call sites benefit from the same fix (Increase Deposit, Withdraw, Stop Loss, Take Profit, Open Position, etc.) — so this is a single shared fix with broad impact.

## Steps to Reproduce

1. Open the dashboard at \`/dashboard\` with a wallet connected on a chain different from the position's chain (e.g. position is on Ethereum mainnet, wallet on Arbitrum or Sepolia).
2. Click "Collect Fees" on a tokenized Uniswap V3 position.
3. Observe the modal: "Wrong network. Please switch to Ethereum Mainnet" + "Switch" button appears.
4. Next to it, the "Collect" button is rendered in the same blue primary style as when on the correct chain. Inspect: \`disabled: false\`, \`pointer-events: auto\`, \`opacity: 1\`.
5. Clicking "Collect" is silently ignored (no UI feedback).

## Suggested Fix

Extend the \`showButtons\` logic or directly add disabled attributes in \`EvmTransactionPrompt\`:

\`\`\`tsx
<button
  onClick={isError ? handleRetry : handleExecute}
  disabled={!enabled && !isError}
  className=\"px-4 py-1.5 bg-blue-600 text-white text-sm rounded-lg font-medium hover:bg-blue-700 disabled:opacity-50 disabled:cursor-not-allowed disabled:hover:bg-blue-600 transition-colors cursor-pointer\"
>
\`\`\`

(Keep retry clickable even when \`enabled: false\`, since retry is meant for error recovery.)

0xNedAlbo:
The transaction prompt's action button rendered in full blue primary styling regardless of `enabled`. `handleExecute` gated on `enabled` and returned early, so on a wrong network the click was silently swallowed — no `disabled` attribute, no cursor change, no feedback of any kind.

Three commits, in dependency order.

### 1. Decouple the status subscription from `enabled` (Q3)

The precondition for everything else. `enabled` gated two unrelated questions: may the user *start* a transaction, and should we *follow* one already submitted.

`useWatchTransactionStatus` consumes `enabled` twice — the effect that creates the subscription (`if (!enabled || !txHash) return;`) and the poll query (`enabled: enabled && !!subscriptionId`). When `enabled` flips false mid-flight, polling stops while the last poll result survives in the query cache (`gcTime` 5 min). `EvmTransactionPrompt` then reads `txWatch.status === 'pending'` forever and renders the confirming spinner for a transaction that has long since succeeded.

That is not hypothetical here: `canCollect` already contains `!isWrongNetwork`, so switching networks during a collect freezes the spinner on `main` today. Making the button track `enabled` without fixing this would have traded one silent failure for another.

The subscription is now gated on the tx hash alone. A hash only exists after a submission `handleExecute` already gated, so nothing can start watching that was not allowed to execute — `enabled` was contributing nothing to *starting* the watch, only the ability to stop it mid-flight, which was the defect. `useWatchTransactionStatus` itself is unchanged; its `enabled` option is a reasonable general-purpose control, and the bug was in how the prompt fed it.

Side benefit: the creation effect no longer re-runs on every `enabled` toggle, which previously re-POSTed a subscription for the same hash each time it flipped back to true.

### 2. Disable the button

```tsx
disabled={isButtonDisabled}
className="... disabled:opacity-50 disabled:cursor-not-allowed disabled:hover:bg-blue-600 ..."
```

- `isButtonDisabled = !enabled && !isError` — retry stays reachable. `handleRetry` only resets local state and is the way *out* of an error, not a way to execute.
- `disabled:opacity-50 disabled:cursor-not-allowed` is the established convention in this app (77 uses each). `disabled:hover:bg-blue-600` is needed because `:hover` still matches a disabled button.
- `showButtons` untouched. Folding `enabled` into it would *hide* the control, removing the affordance and its explanation together.
- `handleExecute`'s `if (!enabled) return;` stays as defence in depth.

### 3. Explain the disabled state (Q2)

New optional `disabledReason` prop, rendered beneath the row while the button is disabled.

Wired at the three forms whose `enabled` predicate *is* the wrong-network check. All three render the prompt only once connected and on the owner account, so "not connected" and "wrong account" cannot reach a disabled button — `EvmWalletConnectionPrompt` and `EvmAccountSwitchPrompt` occupy those states and the prompt is not rendered at all. What can reach it:

| Form | Reachable reasons |
| --- | --- |
| `uniswapv3-vault-collect-fees-form.tsx` | wrong network · no unclaimed fees |
| `uniswapv3-collect-fees-form.tsx` | wrong network · no unclaimed fees |
| `uniswapv3-burn-nft-form.tsx` | wrong network |

The zero-fees case had no explanation anywhere in the modal — the "Wrong network" banner only ever covered the network case.

Per the issue discussion, no network-switch affordance was built into `EvmTransactionPrompt`; it does not have one, and `EvmSwitchNetworkPrompt` already sits directly above in all three forms.

### `uniswapv3-burn-nft-form.tsx`

Third instance, not listed in the issue as filed. `canBurn = isConnected && !isWrongNetwork && !isWrongAccount` is literally the wrong-network predicate, so the Burn button had the same defect. Fixed by the same shared change; issue body updated to record it.

### Scope held

The other 13 call sites are untouched — that is the point of fixing the shared component. Reviewed all 16: two render no button at all (`showActionButton: false`), and none relies on the button being clickable while `enabled === false`, so none regresses.

Both withdraw wizards keep their existing `showActionButton: isConnected && !isWrongNetwork && !isWrongAccount`, which *hides* the button instead. Deliberately not converted here — that is a behaviour change riding inside a bug fix. Worth noting for whoever next touches them: now that the disabled state carries a reason, disable-with-reason is the better form and `showActionButton` is the outlier, not an equal convention.

### Verification

- `pnpm --filter @midcurve/ui typecheck` — clean.
- `eslint` could not be run: it fails repo-wide with `TypeError: Converting circular structure to JSON` out of `@eslint/eslintrc`'s config validator, identically on files this branch does not touch. Pre-existing and unrelated; not fixed here.
- **Not yet verified in a browser.** There is no React component test infrastructure in `apps/midcurve-ui` (no `@testing-library/*`, no `jsdom`, no `test` script; `vitest.config.ts` still points at `src/app/api/**`, a path that no longer exists), and the Playwright specs need a live stack and a real wallet, so they cannot drive a wrong-network state. The manual pass over `pnpm dev` — assert `disabled`, opacity, cursor and hover on a position whose chain differs from the wallet's, then switch and assert it re-enables, plus the zero-fees state and one sequential wizard — is outstanding. Say the word and I will run it via the chrome-devtools MCP before you merge.

Closes #42

### DIFF
--- a/apps/midcurve-ui/src/components/common/EvmTransactionPrompt.tsx
+++ b/apps/midcurve-ui/src/components/common/EvmTransactionPrompt.tsx
@@ -46,6 +46,14 @@ export interface EvmTransactionPromptProps {
   enabled?: boolean;
 
   /**
+   * Why the action is unavailable, shown while the button is disabled.
+   * Supply this whenever `enabled` can be false for a reason the surrounding
+   * UI does not already explain — a disabled control with no account of
+   * itself is as opaque as one that silently ignores clicks.
+   */
+  disabledReason?: string;
+
+  /**
    * Whether to show the action button (Start/Retry)
    * Use this to control sequential transaction flow
    */
@@ -165,6 +173,7 @@ export function useEvmTransactionPrompt({
   retryButtonLabel = 'Retry',
   chainId,
   enabled = true,
+  disabledReason,
   showActionButton = true,
   txHash,
   isSubmitting = false,
@@ -275,6 +284,11 @@ export function useEvmTransactionPrompt({
   // Show buttons when idle OR when there's an error (for retry), but not when active
   const showButtons = showActionButton && (isIdle || isError) && !isActive;
 
+  // Retry stays clickable regardless of `enabled` — it only resets local state
+  // and is the way out of an error, not a way to execute.
+  const isButtonDisabled = !enabled && !isError;
+  const showDisabledReason = showButtons && isButtonDisabled && !!disabledReason;
+
   const element = (
     <div
       className={`py-3 px-4 rounded-lg transition-colors ${
@@ -327,13 +341,19 @@ export function useEvmTransactionPrompt({
           {showButtons && (
             <button
               onClick={isError ? handleRetry : handleExecute}
-              className="px-4 py-1.5 bg-blue-600 text-white text-sm rounded-lg font-medium hover:bg-blue-700 transition-colors cursor-pointer"
+              disabled={isButtonDisabled}
+              className="px-4 py-1.5 bg-blue-600 text-white text-sm rounded-lg font-medium hover:bg-blue-700 disabled:opacity-50 disabled:cursor-not-allowed disabled:hover:bg-blue-600 transition-colors cursor-pointer"
             >
               {isError ? retryButtonLabel : buttonLabel}
             </button>
           )}
         </div>
       </div>
+
+      {/* Why the button is disabled */}
+      {showDisabledReason && (
+        <div className="mt-2 pl-8 text-sm text-slate-400">{disabledReason}</div>
+      )}
 
       {/* Error message */}
       {isError && parsedError && (
--- a/apps/midcurve-ui/src/components/common/EvmTransactionPrompt.tsx
+++ b/apps/midcurve-ui/src/components/common/EvmTransactionPrompt.tsx
@@ -182,12 +182,21 @@ export function useEvmTransactionPrompt({
   // Use the provided txHash or internal one
   const effectiveTxHash = txHash || internalTxHash;
 
-  // Watch transaction status via backend subscription
+  // Watch transaction status via backend subscription.
+  //
+  // Deliberately NOT gated on `enabled`. Whether the user may *start* a
+  // transaction depends on the current network and account; whether we follow
+  // one already submitted does not. A transaction confirms on chain no matter
+  // what the wallet is connected to in the meantime, so tying the subscription
+  // to `enabled` would stop tracking mid-flight the moment the user switched
+  // networks — freezing the prompt on a spinner forever. A tx hash only exists
+  // after a submission that `handleExecute` already gated, so the hash itself
+  // is the correct precondition.
   const txWatch = useWatchTransactionStatus({
     txHash: effectiveTxHash ?? null,
     chainId: chainId ?? 1,
     targetConfirmations,
-    enabled: enabled && !!effectiveTxHash && !!chainId,
+    enabled: !!effectiveTxHash && !!chainId,
   });
 
   // Filter user rejection errors
--- a/apps/midcurve-ui/src/components/positions/protocol/uniswapv3-vault/uniswapv3-vault-collect-fees-form.tsx
+++ b/apps/midcurve-ui/src/components/positions/protocol/uniswapv3-vault/uniswapv3-vault-collect-fees-form.tsx
@@ -87,13 +87,13 @@ export function UniswapV3VaultCollectFeesForm({
     !isNotOwner &&
     unclaimedFees > 0n;
 
-  // The prompt only renders once connected and owned by the user, so those two
-  // states never reach a disabled button — these are the reasons that can.
+  // A wrong network is the only reason that can reach a disabled button. The
+  // prompt renders only once connected and owned by the user, and the modal
+  // cannot be opened at all without unclaimed fees — uniswapv3-vault-actions
+  // disables its entry point on `!hasUnclaimedFees`.
   const collectDisabledReason =
     isWrongNetwork && chainConfig
       ? `Switch your wallet to ${chainConfig.name} to collect fees.`
-      : unclaimedFees === 0n
-      ? 'There are no unclaimed fees to collect.'
       : undefined;
 
   const collectFeesTx = useEvmTransactionPrompt({
--- a/apps/midcurve-ui/src/components/positions/protocol/uniswapv3-vault/uniswapv3-vault-collect-fees-form.tsx
+++ b/apps/midcurve-ui/src/components/positions/protocol/uniswapv3-vault/uniswapv3-vault-collect-fees-form.tsx
@@ -87,11 +87,21 @@ export function UniswapV3VaultCollectFeesForm({
     !isNotOwner &&
     unclaimedFees > 0n;
 
+  // The prompt only renders once connected and owned by the user, so those two
+  // states never reach a disabled button — these are the reasons that can.
+  const collectDisabledReason =
+    isWrongNetwork && chainConfig
+      ? `Switch your wallet to ${chainConfig.name} to collect fees.`
+      : unclaimedFees === 0n
+      ? 'There are no unclaimed fees to collect.'
+      : undefined;
+
   const collectFeesTx = useEvmTransactionPrompt({
     label: 'Collect Fees',
     buttonLabel: 'Collect',
     chainId: config.chainId,
     enabled: canCollect,
+    disabledReason: collectDisabledReason,
     txHash: collectFees.collectTxHash,
     isSubmitting: collectFees.isCollecting,
     isWaitingForConfirmation: collectFees.isWaitingForConfirmation,
--- a/apps/midcurve-ui/src/components/positions/protocol/uniswapv3/uniswapv3-burn-nft-form.tsx
+++ b/apps/midcurve-ui/src/components/positions/protocol/uniswapv3/uniswapv3-burn-nft-form.tsx
@@ -54,6 +54,13 @@ export function UniswapV3BurnNftForm({
   );
   const canBurn = isConnected && !isWrongNetwork && !isWrongAccount;
 
+  // The prompt only renders once connected and on the owner account, so a wrong
+  // network is the only reason that can reach a disabled button here.
+  const burnDisabledReason =
+    isWrongNetwork && chainConfig
+      ? `Switch your wallet to ${chainConfig.name} to burn this NFT.`
+      : undefined;
+
   // All hooks must be called before any early returns
   const burnPosition = useBurnPosition({
     tokenId: BigInt(config.nftId),
@@ -74,6 +81,7 @@ export function UniswapV3BurnNftForm({
     buttonLabel: 'Burn',
     chainId: config.chainId,
     enabled: canBurn,
+    disabledReason: burnDisabledReason,
     txHash: burnPosition.burnTxHash,
     isSubmitting: burnPosition.isBurning,
     isWaitingForConfirmation: burnPosition.isWaitingForBurn,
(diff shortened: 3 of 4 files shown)


0xNedAlbo:
## Review

**Q3 is resolved properly, and the argument is what makes it safe rather than the diff.** `effectiveTxHash` can only exist after a submission that already passed the `enabled` gate in `handleExecute`, so the hash strictly implies "execution was permitted". `enabled` could therefore only ever stop the watch, never start it — which means removing it from the watch is not a loosening. That is the reasoning I wanted, and it is checkable, which a claim about intent would not have been.

Your sharpening is right too: this is not a regression the fix would have introduced, it is a defect already live on `main`, because `canCollect` already carries `!isWrongNetwork`. And a frozen spinner is worse than the stale result I imagined — a stale number is wrong, a permanent spinner tells the user their transaction is still in flight when it settled minutes ago.

**Q2 — the narrowing is better than my instruction.** I named three states; only one can reach a disabled button, because the other two are intercepted by separate prompts before this component renders. Wiring the reachable set instead of writing branches nothing can enter is the same discipline that took `erc721` out of the union rather than leaving a case for it. And finding that the genuinely unexplained state was "no fees to collect" — which never had a banner at all — is the better catch, since that one has no upstream prompt covering it.

**Q1 as agreed.** The end-state argument in the PR description rather than an issue is the right weight for it.

### Before merge: the browser pass, and one scenario in particular

A green typecheck says nothing about a change whose entire content is what the user sees. Run it, and make the mid-flight network switch the centrepiece:

Submit a collect on the correct network, switch networks while the transaction is in flight, and confirm the prompt still resolves to success rather than freezing. That is the behaviour Q3 claims to repair, no test covers it, and it is verifiable as a genuine before-and-after because the defect is live on `main` today. If it resolves after the switch, the argument above is confirmed by the system rather than by reading.

Second: the disabled button on the wrong network states its reason, and the "no fees to collect" case now explains itself.

Report what you saw, including anything that looked off in passing.

### The ESLint failure belongs on #92

This is new information there, and it sharpens the question I asked on that issue. I wanted to know whether `pnpm lint` was red and ignored, or never invoked at all. The answer turns out to be worse than either: it does not reach linting. `@eslint/eslintrc` throws `Converting circular structure to JSON` while validating config, repo-wide, on files this branch never touched — so nothing in this repository has been linted for some time, and the failure does not look like a lint failure to anyone who runs it.

Add it there with the error text. Not for this PR.

0xNedAlbo:
## Browser pass — run, with one gap and one finding that changed the code

Ran against `pnpm dev` on the branch, Chrome on CDP, real Arbitrum mainnet. The repo has a `midcurve-dev-burner` connector (`burner-connector.ts`) that signs in-browser, so SIWE, `eth_sendTransaction` and `switchChain` were all drivable without your wallet — the signature limitation I assumed in the plan does not apply here.

Test subject: WETH/USDC vault position on Arbitrum, `0x7eBA…FEe8`, owned by the burner, 0.639 USDC unclaimed. Wallet started on Sepolia.

### Confirmed

**Wrong network — the reported bug is fixed.**

```json
{"label":"Collect","disabled":true,"opacity":"0.5","cursor":"not-allowed",
 "promptBoxText":"Collect Fees\nCollect\nSwitch your wallet to Arbitrum One to collect fees."}
```

Both the banner ("Wrong network. Please switch to Arbitrum One" + Switch) and the disabled button with its own reason line render together. Screenshot confirms the greyed button.

**Re-enables on switch.** After clicking Switch → chainId 42161: `disabled: false`, `opacity: "1"`, `cursor: "pointer"`, reason line gone.

**Full transaction path works.** Clicked Collect, burner signed, tx landed. Unclaimed fees 0.639 → 0.00, Total PnL −1.46 → −0.829, prompt resolved to success. The decoupled subscription tracked it end to end: `POST …/status/watch` → 202, then `GET …/status/watch/{id}` → `{"txStatus":"success","confirmations":3,"isComplete":true}`. So commit 1 did not break normal tracking.

### The mid-flight switch: not exercised, and I could not force it

This is the one you asked to be the centrepiece, and I did not get it.

Arbitrum confirmed the collect in **~2 seconds** — the watch endpoint already reported `confirmations: 3, isComplete: true` at t=1990ms. My switch call could not land inside that window; by the time it would have, the prompt had already resolved. Then the fees were zero, so there was no second collect to run.

I could not manufacture another attempt without doing something I would not do unprompted: Withdraw and Increase Deposit both move real funds, and the only slow-block position (PAXG/USDC on Ethereum) is owned by `0xDA03…dEd7`, not the burner, so its prompt never renders.

So the Q3 claim is confirmed by argument and by the happy path, **not** by the before-and-after you wanted. To actually close it, one of:

- a position on Ethereum owned by the test wallet — 12s blocks give a workable window;
- temporarily raising `targetConfirmations` on the collect prompt to hold it in `confirming` long enough to switch, then reverting.

Say which and I will run it. I would rather leave this stated as an open gap than let the happy path stand in for it.

### Finding: the "no fees" reason was unreachable — removed (`34317431`)

The zero-fees case I added in the previous commit can never render, and the browser pass is how I found out. With fees at zero I went to reopen the modal and could not: the card's own Collect Fees button is disabled.

Both entry points already gate on it — [`uniswapv3-actions.tsx:191`](apps/midcurve-ui/src/components/positions/protocol/uniswapv3/uniswapv3-actions.tsx#L191) and [`uniswapv3-vault-actions.tsx:147`](apps/midcurve-ui/src/components/positions/protocol/uniswapv3-vault/uniswapv3-vault-actions.tsx#L147), both `disabled={!hasUnclaimedFees}` — and neither position detail page offers a collect action at all (only tabs). No fees, no modal, so the form behind it never renders.

Removed from both collect forms. A wrong network is now the only reason that reaches a disabled button in all three, matching the burn form. Wiring only the reachable set was the point of the earlier commit; this branch turned out not to be in it, and I only learned that by trying to look at it.

Note the wrong-network path is byte-identical after this commit — the verified behaviour above is unaffected.

### Noticed in passing

**`disabledReason` was already a convention here.** [`uniswapv3-vault-actions.tsx:185,208`](apps/midcurve-ui/src/components/positions/protocol/uniswapv3-vault/uniswapv3-vault-actions.tsx#L185) pass `disabled` + `disabledReason` to `VaultStopLossButton` / `VaultTakeProfitButton` in exactly this pairing. I picked the name independently; it matching is luck, but it does mean the prop reads as house style rather than a new invention.

**Two subscriptions per transaction, and a 400.** One collect produced two `POST …/status/watch` (202) with two distinct subscription IDs about 1ms apart, both then polled — consistent with React StrictMode double-invoking the effect in dev. A third POST returned **400**, with no corresponding entry in `logs/dev.log`. Neither affected the outcome. I did not chase it: it is unrelated to this change and predates it. Worth its own issue if you want — say so and I will file it with the request log.

### Checks

- `pnpm --filter @midcurve/ui typecheck` — clean, after the removal.
- ESLint still cannot run; filed on #92 with the trace ([comment](https://github.com/0xNedAlbo/midcurve-finance/issues/92#issuecomment-5232289694)). It turns out the UI's `eslint.config.mjs` still extends `next/core-web-vitals` via `FlatCompat` on an app that is no longer Next.js, and `eslint-config-next` is not even a declared dependency — so the "only the UI is actually linted" conclusion in #92 was wrong; nothing is.

Not merged, issue still open.

PR #110, opened 2026-08-09
fix: current position state from the exact pool price, and one read for both tabs

0xNedAlbo:
Two independent defects made the Overview and Conversion tabs disagree about what a position currently holds. Only one of them is the defect #32 was filed under, and it is the smaller one.

Closes #32

## Three commits, described separately on purpose

**1. `fix(ui)` — the current state from the exact `sqrtPriceX96`.** `calculatePositionStateAtTick()` took a tick and converted it back to a sqrt price. For the range boundaries that is exact, since a boundary *is* a tick. For the current price it is not: `currentTick` is the floor of the tick implied by `sqrtPriceX96`, so the round trip discards the sub-tick remainder — worth 0.013–0.094 % of each leg across the six positions in the dev database.

Split into a sqrt-price core with two callers, and the displayed pool price now comes from the same sqrt price as the amounts. The tick-equality test that decided whether unclaimed fees count became an explicit flag, which also fixes a latent case: a position whose `tickLower` or `tickUpper` equalled `currentTick` added unclaimed fees to the range-boundary PnL.

**2. `fix(ui)` — break-even on a continuous price.** A second defect in the same file, [flagged while planning](https://github.com/0xNedAlbo/midcurve-finance/issues/32#issuecomment-5232406897) and [folded in per the decisions](https://github.com/0xNedAlbo/midcurve-finance/issues/32#issuecomment-5232433261) — not part of the first fix. `calculateBreakEvenPrice()` bisects on price but evaluated candidates through `priceToTick()`, which snaps to `tickSpacing`, sampling position value on a grid quantized to 0.6 % (spacing 60) or 0.1 % (spacing 10). Now converts price → sqrt directly, as `UniswapV3Position.simulatePnLAtPrice()` already does for the PnL curve.

**3. `fix` — one pool read feeds both tabs.** The defect that actually produces gaps of the reported size. The Conversion tab's query holds the server's response for 60 s with no refetch interval, and that response came from an independent read of the stored pool state.

The confirmation run turned up why that framing was wrong, and made the fix smaller: `uniswapv3-position-detail.tsx:43` and its vault twin already wrap the position in `useUniswapV3LiveMetrics`, which watches the pool over a WebSocket-backed subscription and patches the live `sqrtPriceX96` into `position.pool.state` before handing it to every tab. **The single read already existed; the Conversion tab was the one consumer discarding it.** So no new API parameter and no second poller — the price-dependent tail of `computeUniswapV3ConversionSummary()` is split into `applyCurrentPrice()`, and both conversion tabs re-apply it to the price they already hold. The replay over ledger history is price-independent; every event carries the `sqrtPriceX96` it happened at.

`computeUniswapV3ConversionSummary()` keeps its signature and calls the new function with the stored price, so the REST route and the MCP `get_position_conversion` tool are unchanged.

## Verification

`pnpm typecheck` 14/14 · `pnpm build` 11/11 · all six CI suites green (shared 142, api-shared 22, mcp-server 13, services 156, business-logic 24, onchain-data 7). Eight of those tests are new.

**Both defects were confirmed in a browser before the fix**, on the live Arbitrum vault position `0x7eBA…` with the dev stack running.

The precision defect was already user-visible, which I had not expected at 0.01 %. Before:

```
Current Value    20.35 USDC     ← backend, exact sqrtPriceX96 → 20.349689
Position Value   20.34 USDC     ← this file, tick-derived      → 20.349268
```

Same card, one cent apart, a 0.002 % difference straddling a rounding boundary. After the fix both read 20.34. Recomputing against the exact `sqrtPriceX96` the page was serving confirmed the conversion endpoint's `currentBase`/`currentQuote` match the exact branch to the wei, so the Overview was the odd one out. Break-even moved 2,023.37 → 2,024.29 USDC, about 0.045 % — the expected order for a half tick-spacing step on a spacing-10 pool.

The staleness mechanism was confirmed and its magnitude was not. Over ~70 s: 15 fetches of the position endpoint against **one** of the conversion endpoint, with the tab still rendering a 15:59:59 response after the pool state had been re-read at 16:00:23. The dev pool was quiet enough that both reads carried the same price, so nothing visibly diverged in that window. I did not reproduce the 34 % from the report, only the machinery that produces it when the pool moves.

## What is unverified by any suite

The new tests cover `applyCurrentPrice()` in `@midcurve/shared`, where a runner exists — per the [#100 precedent](https://github.com/0xNedAlbo/midcurve-finance/pull/100), and deliberately **not** by standing up UI test infrastructure, which belongs to #91. They pin the arithmetic, idempotence, closed-position pricing, fee attribution, and that the current price is not routed through a tick.

What no suite covers is the UI wiring itself — that the two conversion tabs and the overview actually pass the live price in. `apps/midcurve-ui` has no working unit test runner: `vitest.config.ts` includes only `src/app/api/**/*.e2e.test.ts`, `src/app/` no longer exists, there is no `vitest` devDependency and no `test` script, and CI runs none of it. The existing `src/lib/__tests__/update-position-in-list-cache.test.ts` has never executed. That wiring is browser-verified only. Flagging rather than claiming it.

## Things worth stating rather than leaving implicit

**Displayed numbers move for users**, by up to ~0.1 % on the Current State card and ~0.05 % on break-even. "Position Value" should now equal the "Current Value" card to the wei for NFT positions — same liquidity, same price, same function — which is the cleanest way to check the fix landed. For vaults it stays proportional by design.

**The vault conversion tab changes liquidity basis.** It now uses the holder's proportional liquidity (`liquidity × sharesBalance / totalSupply`), matching the Overview. The vault route feeds `sharesBalance` in directly, which agrees only while `totalSupply` equals the vault's liquidity — true for all three vaults in the dev database today, so this closes a latent divergence rather than a live one.

**`applyCurrentPrice()` is idempotent by construction**, because the UI applies it to a summary the server already priced. It rebuilds each segment's `avgPrice` from that segment's own `deltaBase`/`deltaQuote` and *sets* rather than adds the trailing segment's `feesEarned`. Pinned by two tests, including one that applies a second, different price.

**`ConversionSummary` gains `closePriceX96`.** A closed position prices at close, not at the live price, and that reference price used to be recovered by scanning the ledger inside the replay. Carrying it on the summary is what lets the price step be applied by a caller that never saw the events.

**The issue body was corrected before implementing**, following the #42 pattern: the reported example numbers cannot be produced by the defect the issue diagnosed, and a reproduction step credited tick rounding with the whole gap.

**Not fixed here:** the mini PnL curve, portfolio simulator and range status line still derive the current price from `currentTick`. Same defect class, but they position chart markers where a 0.01 % error is far below a pixel. Also untouched: the in-range badge's inclusive upper bound, which disagrees with the amounts math at exactly `currentTick === tickUpper`.

**Pre-existing and unrelated:** `pnpm lint` in `apps/midcurve-ui` fails to load its eslint config (circular structure in `plugins.react`). It reproduces on a clean tree without these changes, and CI does not run lint.

### DIFF
--- a/apps/midcurve-ui/src/components/positions/protocol/uniswapv3-vault/uniswapv3-vault-conversion-tab.tsx
+++ b/apps/midcurve-ui/src/components/positions/protocol/uniswapv3-vault/uniswapv3-vault-conversion-tab.tsx
@@ -1,7 +1,13 @@
 "use client";
 
+import { useMemo } from "react";
+import { applyCurrentPrice } from "@midcurve/shared";
 import type { UniswapV3VaultPositionData } from "@/hooks/positions/uniswapv3-vault/useUniswapV3VaultPosition";
-import type { UniswapV3VaultPositionConfigResponse } from "@midcurve/api-shared";
+import type {
+  UniswapV3VaultPositionConfigResponse,
+  UniswapV3VaultPositionStateResponse,
+  UniswapV3PoolStateResponse,
+} from "@midcurve/api-shared";
 import { ConversionSummary } from "../uniswapv3/conversion-summary";
 import { useUniswapV3VaultConversion } from "@/hooks/positions/uniswapv3-vault/useUniswapV3VaultConversion";
 
@@ -18,5 +24,32 @@ export function UniswapV3VaultConversionTab({ position }: UniswapV3VaultConversi
     config.ownerAddress,
   );
 
-  return <ConversionSummary summary={summary ?? null} isLoading={isLoading} />;
+  // Re-apply the pool price the rest of the page is rendering — see the NFT tab
+  // for why. The liquidity basis is the holder's proportional share, the same
+  // one the Overview tab uses; the route derives it from sharesBalance alone,
+  // which only agrees while totalSupply equals the vault's liquidity.
+  const pricedSummary = useMemo(() => {
+    if (!summary) return null;
+
+    const state = position.state as UniswapV3VaultPositionStateResponse;
+    const poolState = position.pool.state as UniswapV3PoolStateResponse;
+
+    const totalSupply = BigInt(state.totalSupply);
+    const userLiquidity =
+      totalSupply > 0n
+        ? (BigInt(state.liquidity) * BigInt(state.sharesBalance)) / totalSupply
+        : 0n;
+
+    return applyCurrentPrice(summary, {
+      isToken0Quote: position.isToken0Quote,
+      tickLower: config.tickLower,
+      tickUpper: config.tickUpper,
+      sqrtPriceX96: BigInt(poolState.sqrtPriceX96),
+      liquidity: userLiquidity,
+      unclaimedFees0: BigInt(state.unclaimedFees0),
+      unclaimedFees1: BigInt(state.unclaimedFees1),
+    });
+  }, [summary, position, config.tickLower, config.tickUpper]);
+
+  return <ConversionSummary summary={pricedSummary} isLoading={isLoading} />;
 }
--- a/apps/midcurve-ui/src/components/positions/protocol/uniswapv3/uniswapv3-conversion-tab.tsx
+++ b/apps/midcurve-ui/src/components/positions/protocol/uniswapv3/uniswapv3-conversion-tab.tsx
@@ -1,6 +1,13 @@
 "use client";
 
+import { useMemo } from "react";
+import { applyCurrentPrice } from "@midcurve/shared";
 import type { UniswapV3PositionData } from "@/hooks/positions/uniswapv3/useUniswapV3Position";
+import type {
+  UniswapV3PositionConfigResponse,
+  UniswapV3PositionStateResponse,
+  UniswapV3PoolStateResponse,
+} from "@midcurve/api-shared";
 import { ConversionSummary } from "./conversion-summary";
 import { useUniswapV3Conversion } from "@/hooks/positions/uniswapv3/useUniswapV3Conversion";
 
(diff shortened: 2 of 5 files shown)


0xNedAlbo:
## Review

**The bar I set was already met, and finding that out is the better fix.** I asked for one read feeding both views. `useUniswapV3LiveMetrics` was already doing exactly that — watching the pool and patching live `sqrtPriceX96` into `position.pool.state` before any tab sees it. The defect was not two pollers needing reconciliation; it was one consumer discarding a read it already had.

That distinction matters for the shape of the result. Reconciling pollers would have added coordination to a system that did not need it, and the coordination would have looked like the fix while the discard stayed. Re-applying the price both tabs already hold removes the defect without adding a moving part — no API parameter, no second poller, REST and MCP untouched.

**The precision defect being user-visible is the more interesting finding.** 0.002 % straddling a rounding boundary put 20.35 and 20.34 on the same card. It is worth saying what that generalises to: for a displayed value, "the error is far below anything anyone could notice" is not a safe inference. Rounding boundaries turn arbitrarily small errors into visibly different numbers, and a card showing one quantity twice with two answers costs a user their confidence in both.

**The unreproduced magnitude, reported as unreproduced, is the right handling.** The mechanism is confirmed — fifteen position fetches against one conversion fetch, the tab rendering a response computed before the pool was re-read. The 34 % is not, because the dev pool was quiet enough that both reads carried the same price. Dressing that up as a repro would have been easy and would have hidden the one thing still open.

It also leaves a falsifiable prediction, which is worth putting in the closing comment: a 60-second `staleTime` during a volatile stretch produces exactly the 1–2 % price gap the reported shift needs, so the mechanism plausibly explains the magnitude under conditions dev did not have. **If Jan sees a discrepancy of that size again after this lands, there is a third cause and this issue was closed too early.** Stating that now is worth more than an assurance.

**The vault conversion tab change needs to be visible, not just flagged.** Using the holder's proportional liquidity is correct and consistent with the Overview, but it changes numbers on a screen Jan reads — that belongs in the closing comment in its own right, not only in a PR body that stops being read after the merge.

Counts reconcile: `shared` 134 → 142 with eight new, everything else unchanged, and the arithmetic has now held across five consecutive changes. Testing the extracted arithmetic in `@midcurve/shared` rather than standing up UI infrastructure follows #100 and keeps #91's decision where it belongs.

Approved to merge.

